PR #76, opened 2019-08-15
Add Blurb on Including CPM

DeveloperPaul123:
Added demo cmake code to auto download CPM from github.

This is currently how I include CPM in my projects and find it to be really convenient. Thought I would share. If you are willing to include this let me know what you'd like changed (if anything). 

Thanks for the awesome project!

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,22 @@ For everything else, a target can be created manually (see below).
 
 ## Usage
 
+### Including CPM
+The most simple way to include CPM is to download `CPM.cmake` directory and add it to your project. 
+
+Another way to do this is to have CMAKE download it for you:
+
+````cmake
+set(CPM_DOWNLOAD_LOCATION "${CMAKE_BINARY_DIR}/cmake/CPM.cmake")
+
+if(NOT EXISTS ${CPM_DOWNLOAD_LOCATION})
+	message(STATUS "Downloading CPM.cmake")
+	file(DOWNLOAD https://raw.githubusercontent.com/TheLartians/CPM/master/cmake/CPM.cmake ${CPM_DOWNLOAD_LOCATION})
+endif(NOT EXISTS ${CPM_DOWNLOAD_LOCATION})
+
+include(${CPM_DOWNLOAD_LOCATION})
+````
+
 After `CPM.cmake` has been added to your project, the function `CPMAddPackage` can be used to fetch and configure a dependency.
 Afterwards, any targets defined in the dependency can be used directly.
 `CPMAddPackage` takes the following named parameters.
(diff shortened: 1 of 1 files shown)


TheLartians:
Hey DeveloperPaul123, thank you for your contribution! It does make sense to download CPM directly instead of bundling it in the source.

- The script always downloads the master version of CPM. This is dangerous as it could break your project if CPM behaviour changes in the future. A better way is to add the current used version explicitly: `https://raw.githubusercontent.com/TheLartians/CPM/v0.12/cmake/CPM.cmake`.
- I feel this would fit better into the [Adding CPM](https://github.com/TheLartians/CPM#adding-cpm) section as it also contains the commands to download CPM via bash. Perhaps we could link from the Usage section to there.
- The [Updating CPM](https://github.com/TheLartians/CPM#updating-cpm) section below would need to updated as well (simply change the version number in the URL).

Maybe it would be best to add a custom [Wiki page](https://github.com/TheLartians/CPM/wiki) on downloading CPM via CMake and referring to there from the [Adding CPM](https://github.com/TheLartians/CPM#adding-cpm) section? That would also make it easier to keep the version in the download script up to date.

DeveloperPaul123:
@TheLartians 

Good points, especially with using the current version. I think moving forward it would be best to move the detailed information to a wiki page and refer to that page in the README. Any suggestions on what the wiki page could be called? Does "Adding CPM" work?

TheLartians:
Sounds good! "Adding CPM" works for me.

DeveloperPaul123:
Below is a draft of the wiki page I have. What do you think?


---
As outlined in the README, there is more than one way to include `CPM.cmake` in your project. In addition to using the command line, you can add CMake code to auto download a specific release of `CPM.cmake` to your project an include it automatically.

````cmake
set(CPM_DOWNLOAD_LOCATION "${CMAKE_BINARY_DIR}/cmake/CPM.cmake")
set(CPM_VERSION 0.12)

if(NOT EXISTS ${CPM_DOWNLOAD_LOCATION})
	message(STATUS "Downloading CPM.cmake")
	file(DOWNLOAD https://raw.githubusercontent.com/TheLartians/CPM/v${CPM_VERSION}/cmake/CPM.cmake ${CPM_DOWNLOAD_LOCATION})
endif(NOT EXISTS ${CPM_DOWNLOAD_LOCATION})

include(${CPM_DOWNLOAD_LOCATION})
````
To update the version of CPM that is included, just update the `CPM_VERSION` variable with the correct version *number*, omitting the `v` prefix.

TheLartians:
Looks good to me. One minor thing is that I like to indent using two spaces, so that could be used to stay consistent in the documentation.

DeveloperPaul123:
Wiki page is [live](https://github.com/TheLartians/CPM/wiki/Adding-CPM)

PR #84, opened 2019-10-08
Added GTest (googletest) Example

DeveloperPaul123:
The examples tests the Fibonacci library similar to the doctest example.

### DIFF
--- a/examples/gtest/CMakeLists.txt
+++ b/examples/gtest/CMakeLists.txt
@@ -1,5 +1,7 @@
 cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
 
+project(CPMExampleGtest)
+
 # ---- Options ----
 
 option(ENABLE_TEST_COVERAGE "Enable test coverage" OFF)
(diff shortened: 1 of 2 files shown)


examples/gtest/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
+cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
+
```
TheLartians: This silences the CMake warning about the missing `project()` command. I'll also [add it](https://github.com/TheLartians/CPM/pull/83/commits/4afc4f09d59857bda8c40d9e19dff6eba844d812) to the other examples with for the next version.

```suggestion

project(CPMExampleGtest)

```

DeveloperPaul123:
I wasn't sure about the `project()` because it wasn't in the other examples. But makes sense. 

And yes, I will update the wiki as well once this is merged.

DeveloperPaul123:
I went ahead and updated the wiki.

PR #100, opened 2020-03-16
Add fmt example and document version from git tag

### LINKED ISSUE #98: load fmt as external fail
the tag of [fmt](https://github.com/fmtlib/fmt.git) has no v prefix
```
6.1.0
6.1.1
6.1.2
qls
v0.11.0
```
default strategy will add v prefix to VERSION supplying by user.
```
148   if (NOT DEFINED CPM_ARGS_GIT_TAG) 
149     set(CPM_ARGS_GIT_TAG v${CPM_ARGS_VERSION}) 
150   endif()
```
should we give an option to handle this situation as well as VERSION equal master?

TheLartians:
## Update

As can be seen below the Idea of adding a `VERSION_PREFIX` argument has been dropped due to lacking CMake support for passing empty strings as arguments. Instead, the documentation has been updated and fmt has been added as an example project with prefix-less git tags.

## Original PR

This adds a more consistent syntax for adding packages with non `v`-prefixed git tags. Example:

```cmake
CPMAddPackage(
  NAME fmt
  VERSION 6.1.2
  VERSION_PREFIX "" # fmt uses non-prefixed git tags
  GITHUB_REPOSITORY fmtlib/fmt
)
```

closes #98.


TheLartians:
Turns out in pre 2019 cmake versions there is [no way to detect empty arguments](https://gitlab.kitware.com/cmake/cmake/issues/18706), which is why setting `VERSION_PREFIX ""` won't be recognised.

PR #111, opened 2020-04-26
Fixed link in README.md

GiuseppeCesarano:
The section `Adding CPM` was poiting to non exinsting wiki page


TheLartians:
Ah thanks for noticing!

PR #114, opened 2020-04-30
Fix wiki link to preparing projects for CPM.cmake

nightlark:
- Fixes the wiki link in the README to the page on preparing projects for CPM.cmake
- Adds missing `c` to `ccache`


TheLartians:
Looks good, thanks for noticing!

PR #116, opened 2020-05-09
Use STREQUAL rather than MATCHES <regex>

johelegp:
My directory includes `c++`, which was being interpreted as a regex, as documented at https://cmake.org/cmake/help/latest/command/if.html#condition-syntax.


TheLartians:
Oh good point, that was definitely meant to be `STREQUAL`, thanks for fixing!

Issue #129, opened 2020-06-11
[Feature Request] Use shallow clone for git repositories


kingsamchen:
CPM currently clones full history for git-repository dependencies, which takes more time than necessary (in most cases) on downloading, and consumes more disk space.

Maybe use git shallow clone _by default_ for those dependencies?

There is a `GIT_SHALLOW` option to enable `--depth=1` with clone, which does what we exactly desire.

If I may, I would like to make a PR on it, :-)

TheLartians:
Hey thanks for the suggestion and PR! I agree that it would be great to have `GIT_SHALLOW` enabled by default! However, as you can see from the CI failure, enabling `GIT_SHALLOW` can lead to unexpected failures as CMake fails to checkout certain branches or commits for some reason.

Until that is resolved, it's therefore currently safer to enable `GIT_SHALLOW` on a per-dependency basis, knowing when CMake will be able to successfully check out the commit. You can activate it by providing the `GIT_SHALLOW ON` argument together with the dependency source parameters in `CPMAddPackage`.

kingsamchen:
According to this answer https://stackoverflow.com/questions/31278902/how-to-shallow-clone-a-specific-commit-with-depth-1 clone with `depth 1` won't work on a specific commit HASH. That is why the yaml example failed.

Maybe making `GIT_SHALLOW` a per-dependency option is a more practical approach.

TheLartians:
> According to this answer https://stackoverflow.com/questions/31278902/how-to-shallow-clone-a-specific-commit-with-depth-1 clone with depth 1 won't work on a specific commit HASH. That is why the yaml example failed.

Shallow fetching a real git tag seems to work though, perhaps we can set a `SHALLOW` default behaviour when we detect the `GIT_TAG` to be an actual tag as opposed to a hash.

kingsamchen:
A wonderful idea.

And maybe it is easier to identify out a hash from tag/branch value, because commit hash has fixed length and contains no characters other than hex digits.

While tags from different projects could use varied formats.

E.g for [standalone ASIO](https://github.com/chriskohlhoff/asio/releases), its release tag is in the form of `asio-1-23-4`; it doesn't even have prefix `v`, and yet uses `-` instead of conventional `.` as the version delimiter.

kingsamchen:
@TheLartians Hi, I just made another commit on enabling shallow clone conditionally. Could you please review the commits when you are free?

Any suggestion is welcome

[...]

PR #130, opened 2020-06-11
Use shallow clone for git repositories by default (#129)

kingsamchen:
* use shallow clone for git repositories by default

* remove trailing spaces


cmake/CPM.cmake
```diff
@@ -466,11 +480,25 @@ endfunction()
...
   else()
     string(REGEX MATCH "v?([0123456789.]*).*" _ ${GIT_TAG})
     SET(${RESULT} ${CMAKE_MATCH_1} PARENT_SCOPE)
   endif()
 endfunction()
+
+# guesses if the git tag is a commit hash or an actual tag or a branch nane.
+function(cpm_is_git_tag_commit_hash GIT_TAG RESULT)
+  string(LENGTH ${GIT_TAG} length)
+  if (NOT (length EQUAL 40))
```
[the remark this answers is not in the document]
kingsamchen: Thanks for the tip! and speaking of short hash, any length between `[7, 40)` could be a legal short hash, though 7-character is not enough for few super-big projects.

Should we treat lengths within the entire range as possible candidates of a commit hash ? given the guarantee of no commit hash should be mis-detected. 

Ref: https://stackoverflow.com/questions/18134627/how-much-of-a-git-sha-is-generally-considered-necessary-to-uniquely-identify-a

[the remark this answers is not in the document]
TheLartians: Yeah, that's a good point, I agree that we should guarantee that any legal git hash should continue working. And as most version tags would contain a special character or at least a 'v' prefix, we should still be getting shallow clones in almost all cases.

cmake/CPM.cmake
```diff
@@ -230,6 +231,19 @@ function(CPMAddPackage)
...
+
+    # If GIT_SHALLOW is explicitly specified, hornor the value.
+    # Otherwise, enable shallow clone only when GIT_TAG is not a commit hash.
+    # Our guess may not be accurate, but it should guarantee no commit hash get mis-detected.
+    if (DEFINED CPM_ARGS_GIT_SHALLOW)
+      list(APPEND CPM_ARGS_UNPARSED_ARGUMENTS GIT_SHALLOW ${CPM_ARGS_GIT_SHALLOW})
+    else()
+      cpm_is_git_tag_commit_hash(${CPM_ARGS_GIT_TAG} IS_HASH)
+      if (NOT ${IS_HASH})
+        message(STATUS "GIT_TAG=${CPM_ARGS_GIT_TAG} is not a commit hash; enable shallow clone")
```
TheLartians, line 243: I wouldn't print internal config decisions to avoid flooding the output.

kingsamchen: Honestly, that was the first time I didn't use `CPM_SOURCE_CACHE`; and it indeed caused a checkout failure.

I noticed that without global cache, the name of source folder is always `foobar-src`, no version or some tag information entangled, which prevented re-downloading of the dependency; and the old folder is shallow-cloned, therefore resulted in checkout failure.

So I guess your suggestion is necessary to keep the local cache work.

Btw: the logging output is deleted.

cmake/CPM.cmake
```diff
@@ -323,6 +310,18 @@ function(CPMAddPackage)
       list(APPEND CPM_ARGS_UNPARSED_ARGUMENTS DOWNLOAD_COMMAND "${CMAKE_COMMAND}")
       set(PACKAGE_INFO "${download_directory}")
     else()
+      # If GIT_SHALLOW is explicitly specified, hornor the value.
+      # Otherwise, enable shallow clone only when GIT_TAG is not a commit hash.
+      # Our guess may not be accurate, but it should guarantee no commit hash get mis-detected.
+      if (DEFINED CPM_ARGS_GIT_SHALLOW)
+        list(APPEND CPM_ARGS_UNPARSED_ARGUMENTS GIT_SHALLOW ${CPM_ARGS_GIT_SHALLOW})
+      else()
```
TheLartians, line 318: This part should still be considered in the default case (e.g. in line 233), as we still want to allow users to set `GIT_SHALLOW` explicitly in any case.

TheLartians:
Looks great, thanks a lot for the PR! I'll test this in some large projects and will create a new release if everything works as expected.

PR #137, opened 2020-08-03
Fix indeterminate command for GitLab argument

jecassis:
The intended behavior to capture the GitLab argument is better accomplished using `set` rather than `list` as with the GitHub argument above. In addition, there is no documented form of the `list` command with that format.


TheLartians:
Huh, how was that overlooked? We should add a test for GitLab soon.
Thanks for fixing!

TheLartians:
The fix is now live in v0.27.2!

PR #150, opened 2020-09-08
Implemented spdlog library example

leozz37:
Implemented a usage example of [spdlog](https://github.com/gabime/spdlog) library.


TheLartians:
Love spdlog, thanks for the example!

PR #155, opened 2020-09-26
fix passing options when using CPM_<pkg>_SOURCE

germasch:
When I used CPM_<pkg>_SOURCE to point to a locally modified copy, the `OPTIONS` ended up not being passed on. I think they should be, so this patch does pass them to the repeated `CPMAddPackage()`.


TheLartians:
Sounds reasonable. Thanks for the fix!

TheLartians:
The fix is live in v0.27.4!

germasch:
Great, thanks!

PR #159, opened 2020-10-19
fix: GIT_TAG not parsed inside CPMFindPackage

xmuller:
When using CPMFindPackage, if no VERSION is given then GIT_TAG is used as fallback if defined, which is a good idea.
The issue is that GIT_TAG is missing inside the oneValueArgs variable therefore GIT_TAG is not parsed. This is resolved by this little change.


TheLartians:
Good catch, thanks for the fix!

TheLartians:
The fix is live in v0.27.5! 🎉

PR #162, opened 2020-11-17
Feature/prettify package lock

alexandreSalconiDenis:
Hi,
I've added some prettify to the `cpm_add_to_package_lock` and `cpm_add_comment_to_package_lock`.
It should help comparing and editing the package lock files
And will look like example from the Readme.md


TheLartians:
Hey, thanks for the PR! 
I'm a bit hesitant as the package-lock was originally intended to be a small auto-generated file, and not rarely be human read or edited, but I do see that this can become more frequent. Also many projects have automatic code formatting enabled, so maybe instead of accepting the arguments in a string value, we should change the way `CPMDeclarePackage` works to accept the same arguments as `CPMAddPackage`. That way the auto code formatter would take care of any styling and be more consistent with the user's preferences.

cmake/CPM.cmake
```diff
@@ -505,3 +509,62 @@ function(cpm_is_git_tag_commit_hash GIT_TAG RESULT)
     endif()
   endif()
 endfunction()
+
+macro(prettify_cpm_add_package OUT_VAR IS_IN_COMMENT)
```
TheLartians, line 513: Could you change the name to `cpm_prettyfy_package_arguments`?
That way it is more clear and prefixed with `cpm`.

cmake/CPM.cmake
```diff
@@ -505,3 +509,62 @@ function(cpm_is_git_tag_commit_hash GIT_TAG RESULT)
...
+      cmake_parse_arguments(CPM_AGS "" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+      foreach(oneArgName ${oneValueArgs})
+          if(DEFINED CPM_AGS_${oneArgName})
+              if(${IS_IN_COMMENT})
+                  string(APPEND ${OUT_VAR} "#")
+              endif()
+              if(${oneArgName} STREQUAL "SOURCE_DIR")
+                  string(REPLACE ${CMAKE_SOURCE_DIR} "\${CMAKE_SOURCE_DIR}" CPM_ARGS_${oneArgName} ${CPM_ARGS_${oneArgName}} )
+              endif()
```
TheLartians, line 542: Probably we should use `CMAKE_CURRENT_SOURCE_DIR` here, for a more portable approach.

cmake/CPM.cmake
```diff
@@ -505,3 +509,62 @@ function(cpm_is_git_tag_commit_hash GIT_TAG RESULT)
...
+                  if(${IS_IN_COMMENT})
+                      string(APPEND ${OUT_VAR} "#")
+                  endif()
+                  string(APPEND ${OUT_VAR} "        \"${singleOption}\"\n")
+              endforeach()
+          endif()
+      endforeach()
+
+    if(NOT "${CPM_AGS_UNPARSED_ARGUMENTS}" STREQUAL "")
+        string(REPLACE ";" " " CPM_AGS_UNPARSED_ARGUMENTS "${CPM_AGS_UNPARSED_ARGUMENTS}")
```
TheLartians, line 562: This is risky, e.g. what if there's an escaped `;` somewhere in the unparsed arguments? Imo it's better to not touch them.

cmake/CPM.cmake
TheLartians, line 513: Also better use `function` instead of `macro` as it creates a new scope.

test/unit/package-lock_prettify.cmake
```diff
@@ -0,0 +1,67 @@
...
+    SOURCE_DIR \${CMAKE_SOURCE_DIR}/dependency
+    #(unparsed)
+    UPDATE_DISCONNECTED ON TESTCUSTOMDATA TRUE
+)
+")
+
+ASSERT_EQUAL(${READFILE}  ${EXPECTED})
+
+configureWithDeclare(NO)
+execute_process(COMMAND ${CMAKE_COMMAND} -E rm -f ${CMAKE_CURRENT_LIST_DIR}/local_dependency/package-lock_prettify.cmake)
```
TheLartians, line 67: You probably mean `${CMAKE_CURRENT_LIST_DIR}/local_dependency/package-lock.cmake`

test/unit/package-lock_prettify.cmake
```diff
@@ -0,0 +1,67 @@
+
```
TheLartians, line 1: I'd prefer a more simple unit test of the output of `cpm_prettyfy_package_arguments()` here, as the integrated functionality is already tested. Optimally with multiple input examples.

TheLartians (changes requested):
Thanks for the changes, this looks quite a bit better! I'd like it if we could make the unit test more focused. It should optimally only call and check the output of the prettify function. We already have a unit test that checks the integrated functionality with CMake. Also there still seems to be a merge conflict, could you pull the most recent master and recommit?

cmake/CPM.cmake
alexandreSalconiDenis: Will it work to add in the `CPMDeclarePackage` some argument with `;` ? 
I tough it would create error if done like this:
```
CPMDeclarePackage(Dependency
      NAME Dependency
      SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/local_dependency/dependency
      UPDATE_DISCONNECTED;ON;TESTCUSTOMDATA;TRUE
    )
```

TheLartians: Hm that is good question, I have no idea how CMake handles mixing spaces and semicolons. The safest might be to loop over `${CPM_AGS_UNPARSED_ARGUMENTS}` and append them one-by-one with a space.

cmake/CPM.cmake
alexandreSalconiDenis: `CMAKE_CURRENT_SOURCE_DIR` will depend on the location where the package was called, `CMAKE_SOURCE_DIR` is the path where is the root `CMakeLists.txt`. Since `SOURCE_DIR` is the location of the package i only resolve the "root path".

alexandreSalconiDenis:
I had some difficulty with the cmake-format but I think now everything is ok, tell me if I there is more to do. I've also change the unit test to cover only the function like you asked.

TheLartians (changes requested):
Thanks for the update, looks good! Just some small things and formatting issues remain.

cmake/CPM.cmake
TheLartians: Hmmmm that is a good point. The source dir will always change depending on if the current package is used as the root or a subdirectory. I can see how this can cause some unexpected behaviour. Maybe we simply shouldn't add any packages using `SOURCE_DIR` to the package lock?

alexandreSalconiDenis: It is an attempt to make the `SOURCE_DIR` more dynamic, since if its not resolve a little bit, the variable will hold the full path of your system, and it will be unusable by other dev. I was hoping to try 1 or 2 term to resolve some part (like if its in `_dep` or under the root of the project) since with that it will be valid for everybody. For us we usually put most package near the root of the project to ease the developing, but I know usually it's under `<build folder>/_dep/...` .

cmake/CPM.cmake
```diff
@@ -625,3 +629,71 @@ function(cpm_is_git_tag_commit_hash GIT_TAG RESULT)
...
+  # set(OUT_VAR "")
+  cmake_parse_arguments(CPM_ARGS "" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+  foreach(oneArgName ${oneValueArgs})
+    if(DEFINED CPM_ARGS_${oneArgName})
+      if(${IS_IN_COMMENT})
+        string(APPEND PRETTY_OUT_VAR "#")
+      endif()
+      if(${oneArgName} STREQUAL "SOURCE_DIR")
+        message("CMAKE_SOURCE_DIR:${CMAKE_SOURCE_DIR}")
```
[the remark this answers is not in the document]
alexandreSalconiDenis: my bad, Indeed I need to remove that.

cmake/CPM.cmake
TheLartians: Yeah, I understand the use case, though this can go wrong when trying to use a package as a dependency that itself has relative imports using `SOURCE_DIR`. In most cases using `SOURCE_DIR` will be for first-party code, so imo we should be fine simply omitting it in that case.

TheLartians:
Oh yeah, fyi: you can easily automatically format all source code by running:

```bash
```

cmake/CPM.cmake
alexandreSalconiDenis: Not sure to understand completely, can you show me the suggestion.
 From what you say we should remove the resolution of the path for the `SOURCE_DIR`, or remove `SOURCE_DIR`?

TheLartians: Never mind, I just saw that we are already ignoring any args, if the `SOURCE_DIR` parameter is specified. Here is the logic that adds packages to the lock:

```cmake
  if(CPM_PACKAGE_LOCK_ENABLED)
    if((CPM_ARGS_VERSION AND NOT CPM_ARGS_SOURCE_DIR) OR CPM_INCLUDE_ALL_IN_PACKAGE_LOCK)
      cpm_add_to_package_lock(${CPM_ARGS_NAME} "${ARGN}")
    elseif(CPM_ARGS_SOURCE_DIR)
      cpm_add_comment_to_package_lock(${CPM_ARGS_NAME} "local directory")
    else()
      cpm_add_comment_to_package_lock(${CPM_ARGS_NAME} "${ARGN}")
    endif()
  endif()
```

Which makes sense, as adding `SOURCE_DIR` to the package lock can introduce unexpected bugs. This also means that the prettify function can happily ignore the `SOURCE_DIR` argument.

alexandreSalconiDenis: ok I will add add the ignore step for `SOURCE_DIR`

TheLartians: To be clear I think we can just delete lines 540-542 as Packages using SOURCE_DIR won't have the argument here anyways.

alexandreSalconiDenis: mmm, could we could simply add a test like:
```
cpm_prettify_package_arguments(PRETTY_ARGN true "local directory")
set(EXPECTED_COMMENTED_LOCALDIR
    "#    local directory
"
)
assert_equal(${PRETTY_ARGN} ${EXPECTED_COMMENTED_LOCALDIR})
```
since if we define: `CPM_INCLUDE_ALL_IN_PACKAGE_LOCK` the `SOURCE_DIR` will be added to the lock file.

alexandreSalconiDenis: this test will cover the `elseif(CPM_ARGS_SOURCE_DIR)` part.

TheLartians: > since if we define: `CPM_INCLUDE_ALL_IN_PACKAGE_LOCK` the `SOURCE_DIR` will be added to the lock file.

Ok gotcha, I wasn't aware that you're explicitly enabling it. As it's disabled by default (it shouldn't be used for dependencies anyways) I guess the original check against `CMAKE_SOURCE_DIR` is fine after all.

cmake/CPM.cmake
```diff
@@ -625,3 +627,69 @@ function(cpm_is_git_tag_commit_hash GIT_TAG RESULT)
...
+      GITLAB_REPOSITORY
+      GIT_REPOSITORY
+      SOURCE_DIR
+      DOWNLOAD_COMMAND
+      FIND_PACKAGE_ARGUMENTS
+      NO_CACHE
+      GIT_SHALLOW
+  )
+  set(multiValueArgs OPTIONS)
+  # set(OUT_VAR "")
```
TheLartians, line 648: Could you remove the comment?
```suggestion
```

TheLartians (commented):
I just tested it on my larger projects and it works great, thanks for the PR and updates! 🚀 
One final request though: the CPM.cmake source code is indented with two spaces, so I think it would make sense for the package lock as well. I think it's currently using four spaces. What do you think?

cmake/CPM.cmake
alexandreSalconiDenis: That's fine I will change the code to work with that format.

TheLartians (approved):
Looks great, thanks!

TheLartians:
The changes are now published in v0.28.2! 🎉

PR #176, opened 2021-01-13
fix: GIT_SHALLOW not used without SOURCE_CACHE

xmuller:
Hello !

When using CPMAddPackage, I noticed a long download time.
After inspection, my conclusion is that the GIT_SHALLOW flag is never add automatically in my case (when using GIT_TAG) because the code is nested inside conditions for CPM_ARGS_UNPARSED_ARGUMENTS.

I think it was not intended to be that way. I rather expect to place the code outside the conditional block, this way GIT_SHALLOW is always ON as long as we download something from git without hash.

What do you think ?


TheLartians:
The original code appears correct, as `GIT_SHALLOW` should only be automatically appended if downloading to a cache directory. Otherwise it is not possible to change the dependency's version without removing the build directory, because CMake doesn't re-fetch a git repository if the tag changes. You can see that in the failed CI tests.

The reason for the poor performance is probably due to CMake's slow FetchContent implementation.

PR #181, opened 2021-01-20
Guess version number from URL

jhasse:
Source tarballs include the version number in the file name, e.g.
http://downloads.xiph.org/releases/ogg/libogg-1.3.4.tar.xz

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -167,13 +167,18 @@ endfunction()
 
 # Find a package locally or fallback to CPMAddPackage
 function(CPMFindPackage)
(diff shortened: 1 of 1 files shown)


TheLartians:
Looks good, thanks for the PR! Some requests:

- Could you also add a unit test in `./tests/unit`, similar to the `version_from_git_tag.cmake` test? This will prevent another PR accidentally breaking the feature.
- Support for GitHub style releases would also be nice. They have a  `https://github.com/<user>/<name>/releases/download/<version_tag>/<filename>` schema.

PR #190, opened 2021-02-04
update most package versions

ClausKlein:
fix  examples with build problems too

see https://github.com/TheLartians/CPM.cmake/issues/189

add `cmake project code injection` as an option when using `build_all.py`

see https://cmake.org/cmake/help/latest/command/project.html#code-injection

tested with Apple clang version 12.0.0 (clang-1200.0.32.29)

### DIFF
--- a/examples/asio-standalone/CMakeLists.txt
+++ b/examples/asio-standalone/CMakeLists.txt
@@ -24,7 +24,7 @@ CPMAddPackage(
 if(asio_ADDED)
   add_library(asio INTERFACE)
 
-  target_include_directories(asio INTERFACE ${asio_SOURCE_DIR}/asio/include)
+  target_include_directories(asio SYSTEM INTERFACE ${asio_SOURCE_DIR}/asio/include)
 
   target_compile_definitions(asio INTERFACE ASIO_STANDALONE ASIO_NO_DEPRECATED)
 
@@ -65,4 +65,4 @@ endif()
 
 add_executable(CPMExampleASIOStandalone main.cpp)
 target_link_libraries(CPMExampleASIOStandalone asio)
-target_compile_features(CPMExampleASIOStandalone PRIVATE cxx_std_20)
+target_compile_features(CPMExampleASIOStandalone PRIVATE cxx_std_17)
--- a/examples/asio-standalone/CMakeLists.txt
+++ b/examples/asio-standalone/CMakeLists.txt
@@ -10,9 +10,9 @@ find_package(Threads REQUIRED)
 
 CPMAddPackage(
   NAME asio
-  VERSION 1.16.1
+  VERSION 1.18.1
   GITHUB_REPOSITORY chriskohlhoff/asio
-  GIT_TAG asio-1-16-1 # asio uses non-standard version tag, we must specify GIT_TAG
+  GIT_TAG asio-1-18-1 # asio uses non-standard version tag, we must specify GIT_TAG
 )
 
 # ASIO doesn't use CMake, we have to configure it manually. Extra notes for using on Windows:
@@ -65,3 +65,4 @@ endif()
 
 add_executable(CPMExampleASIOStandalone main.cpp)
 target_link_libraries(CPMExampleASIOStandalone asio)
+target_compile_features(CPMExampleASIOStandalone PRIVATE cxx_std_20)
--- a/examples/asio-standalone/CMakeLists.txt
+++ b/examples/asio-standalone/CMakeLists.txt
@@ -65,4 +65,4 @@ endif()
 
 add_executable(CPMExampleASIOStandalone main.cpp)
 target_link_libraries(CPMExampleASIOStandalone asio)
-target_compile_features(CPMExampleASIOStandalone PRIVATE cxx_std_17)
+target_compile_features(CPMExampleASIOStandalone PRIVATE cxx_std_11)
--- a/examples/build_all.py
+++ b/examples/build_all.py
@@ -7,7 +7,7 @@ from subprocess import PIPE, run
 
 # NOTE: boost V1.67 is to old! CK
 examples = [
-    x for x in Path(__file__).parent.iterdir() if x.is_dir() and (x / 'CMakeLists.txt').exists() and (not x.name in ['boost'])
+    x for x in Path(__file__).parent.iterdir() if x.is_dir() and (x / 'CMakeLists.txt').exists() and (not x.name in ['boost', 'old-gtest'])
 ]
 
 assert(len(examples) > 0)
@@ -37,8 +37,9 @@ for example in examples:
         before = "-DCMAKE_PROJECT_INCLUDE_BEFORE=%s/before_project_setup.cmake" % (
             cmakeModulesPath)
         after = "-DCMAKE_PROJECT_INCLUDE=%s/build_options.cmake" % (cmakeModulesPath)
-
-    configure = runCommand('cmake -H%s -B%s -G Ninja %s %s' % (example, project, before, after))
+        configure = runCommand('cmake -H%s -B%s -G Ninja %s %s' % (example, project, before, after))
+    else:
+        configure = runCommand('cmake -H%s -B%s' % (example, project))
     print('  ' + '\n  '.join([line for line in configure.split('\n') if 'CPM:' in line]))
     build = runCommand('cmake --build %s -j8' % (project))
     print('  ' + '\n  '.join([line for line in build.split('\n') if 'Built target' in line]))
--- a/examples/build_all.py
+++ b/examples/build_all.py
@@ -5,9 +5,8 @@ import os
 from pathlib import Path
 from subprocess import PIPE, run
 
-# NOTE: boost V1.67 is to old! CK
 examples = [
-  x for x in Path(__file__).parent.iterdir() if x.is_dir() and (x / 'CMakeLists.txt').exists() and (not x.name in ['boost'])
+  x for x in Path(__file__).parent.iterdir() if x.is_dir() and (x / 'CMakeLists.txt').exists()
 ]
 
 assert(len(examples) > 0)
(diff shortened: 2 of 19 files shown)


examples/build_all.py
```diff
@@ -1,29 +1,46 @@
...
 
+import os
+
 from pathlib import Path
 from subprocess import PIPE, run
 
+# NOTE: boost V1.67 is to old! CK
 examples = [
-  x for x in Path(__file__).parent.iterdir() if x.is_dir() and (x / 'CMakeLists.txt').exists()
+    x for x in Path(__file__).parent.iterdir() if x.is_dir() and (x / 'CMakeLists.txt').exists() and (not x.name in ['boost', 'old-gtest'])
```
TheLartians, line 10: On my system (Mac OS 10.15.7, Apple clang 12.0.0) both the boost and gtest examples compile without issues. Which versions are you using and what problems are they creating? Also not sure the `old-gtest` directory exists in the repo.

Tbh I'd rather try to fix the example or remove them completely, if unfixable on modern systems instead of excluding them from the script / CI system.

examples/build_all.py
```diff
@@ -1,29 +1,46 @@
...
-    print("error while running '%s':\n" % command, '  ' + str(result.stderr).replace('\n','\n  '))
-    exit(result.returncode)
-  return result.stdout
+    print('- %s' % command)
+    result = run(command, stdout=PIPE, stderr=PIPE, universal_newlines=True, shell=True)
+    if result.returncode != 0:
+        print("error while running '%s':\n" %
+              command, '  ' + str(result.stderr).replace('\n', '\n  '))
+        exit(result.returncode)
+    return result.stdout
```
TheLartians, line 23: This project uses 2-space indentation throughout for consistency. Could you reformat the python to use 2-space indentation as before?

examples/build_all.py
```diff
@@ -1,29 +1,46 @@
...
+    # https://cmake.org/cmake/help/latest/command/project.html#code-injection
+    #
+    cmakeModulesPath = os.environ['HOME'] + '/Workspace/cmake'
+    if Path(cmakeModulesPath).is_dir():
+        before = "-DCMAKE_PROJECT_INCLUDE_BEFORE=%s/before_project_setup.cmake" % (
+            cmakeModulesPath)
+        after = "-DCMAKE_PROJECT_INCLUDE=%s/build_options.cmake" % (cmakeModulesPath)
+        configure = runCommand('cmake -H%s -B%s -G Ninja %s %s' % (example, project, before, after))
+    else:
+        configure = runCommand('cmake -H%s -B%s' % (example, project))
```
TheLartians, line 42: This seems to be oddly workspace-specific. Is it really necessary?

examples/build_all.py
```diff
@@ -1,29 +1,46 @@
...
+    print('  ' + '\n  '.join([line for line in configure.split('\n') if 'CPM:' in line]))
+    build = runCommand('cmake --build %s -j8' % (project))
```
TheLartians, line 44: Often there aren't many cores available on CI, so adding more threads might actually increase build time.
Suggestion: use the number of available cpu cores available on the system

```suggestion
    build = runCommand('cmake --build %s -j%i' % (project, os.cpu_count()))
```

examples/gtest/CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
 
 project(CPMExampleGtest)
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS NO)
```
TheLartians, line 7: Are these global `set`s actually necessary?

TheLartians (changes requested):
Hey, thanks for the issue and PR! It was about time to give the examples some love and update the dependencies to recent versions.

examples/gtest/CMakeLists.txt
ClausKlein: without this, it does not compile on my host

examples/build_all.py
ClausKlein: yes, some of the creates features of cmake:

I use it to setup the ccache, cxx std , warnings , clang-tidy, inject my modules want to have, ...

examples/build_all.py
ClausKlein: yes, I used autopep8 with it defaults

examples/build_all.py
ClausKlein: it seems, that the build is not independent of the libraries installed on my build host: 

bash-3.2$ brew list
abseil			diffutils		git-flow		jpeg			libtool			openjdk			ruby
adns			docbook			git-gui			jsoncpp			libunistring		openldap		smartmontools
asciidoc		docker			glib			lcov			libusb			openssl@1.1		snappy
astyle			doctest			glog			libassuan		libyaml			p11-kit			source-highlight
autoconf		dos2unix		gmp			libcbor			llvm			pango			spdlog
automake		double-conversion	gnupg			libev			lua			pcre			sqlite
bash-completion		doxygen			gnutls			libevent		lz4			pcre2			swig
bdw-gc			duplicity		go			libffi			lzo			pinentry		tcl-tk
boost			findutils		gobject-introspection	libfido2		make			pixman			tidy-html5
brotli			fmt			googletest		libgcrypt		meson			pkg-config		tree
c-ares			folly			graphite2		libgpg-error		mpfr			poco			unbound
cairo			fontconfig		graphviz		libidn2			multimarkdown		popt			webp
ccache			freetype		grpc			libksba			multitail		protobuf		wget
cereal			fribidi			gts			libmetalink		ncurses			pylint			xmake
cmake			gcc			guile			libmpc			net-snmp		python@3.9		xz
conan			gd			harfbuzz		libpng			netpbm			range-v3		yamllint
coreutils		gdbm			icu4c			librsvg			nettle			rapidjson		zstd
cpp-gsl			gdk-pixbuf		isl			librsync		nghttp2			re2
cppcheck		gettext			jasper			libssh2			ninja			readline
ctags			gflags			jemalloc		libtasn1		npth			renameutils
curl			git			jinja2-cli		libtiff			omniorb			rtmpdump
bash-3.2$

examples/asio-standalone/CMakeLists.txt
```diff
@@ -65,3 +65,4 @@ endif()
 
 add_executable(CPMExampleASIOStandalone main.cpp)
 target_link_libraries(CPMExampleASIOStandalone asio)
+target_compile_features(CPMExampleASIOStandalone PRIVATE cxx_std_17)
```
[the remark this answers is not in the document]
ClausKlein: asio supports all Standard starting with c++11 as far as I know 

on other examples it is done in this way.

c++20 is current std. So 17 IMO should be required and used

examples/build_all.py
ClausKlein: do you have a solution to get boost v1.75?

examples/asio-standalone/CMakeLists.txt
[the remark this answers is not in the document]
TheLartians: Perhaps we should add the `cxx_std_11` flag to asio though, as users could run into issues when linking it to their own library? CMake will automatically enable the most recent standard selected if his library uses a newer version.

examples/build_all.py
TheLartians: Unresolving, as the point above has not been addressed.

> do you have a solution to get boost v1.75?

So far the only CMake compatible version I've seen is [Orphis/boost-cmake](https://github.com/Orphis/boost-cmake). For me and on the CI it hasn't been creating any issues, so I see no problem with keeping it until we can replace it with a newer version.

Also note that `old-gtest` does not seem to exist in this directory, so I don't see the purpose of the exclusion.

examples/build_all.py
TheLartians: I do see the value of this feature, however I'm not sure how it fits into the automated example script here, as there are also many other combinations of arguments that users could potentially pass to CMake that we can't possible all handle explicitly. As a compromise, how about forwarding any extra args to the python script to CMake? Then you could invoke the script using

```bash
./build_all.py -DCMAKE_PROJECT_INCLUDE_BEFORE=my/before/script -DCMAKE_PROJECT_INCLUDE=my/after/script
```

examples/gtest/CMakeLists.txt
TheLartians: It does work on my system and CI, so there is something strange going on here. What version of MacOS, CMake, and clang are you using?

TheLartians (changes requested):
Thanks for updating the formatting! I'd still like the above issues to be addressed before merging.

TheLartians:
As for the apparently system-dependent issues appearing on your setup, could you check if they still arise if you don't use `CMAKE_PROJECT_INCLUDE_BEFORE` and `CMAKE_PROJECT_INCLUDE` to build the examples?

examples/gtest/CMakeLists.txt
ClausKlein: cmake version 3.19.4

Darwin Claus-iMac.fritz.box 20.2.0 Darwin Kernel Version 20.2.0: Wed Dec  2 20:39:59 PST 2020; root:xnu-7195.60.75~1/RELEASE_X86_64 x86_64

I think there are problems with different version of gtest header installed on my host.
I will check this

examples/build_all.py
ClausKlein: good idea ...

I will try it

examples/asio-standalone/CMakeLists.txt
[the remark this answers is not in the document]
ClausKlein: OK

examples/build_all.py
ClausKlein: done

TheLartians (approved):
Looks great, thanks for the PR and updates!

Issue #192, opened 2021-02-06
FetchContent failure when using CPM on multiple projects with cmake < 3.17


xmuller:
##  To reproduce the bug with at least CPM >= 0.27.5:

Minimal structure:
```
.
├── CMakeLists.txt
├── libA
│   ├── CMakeLists.txt
│   └── CPM.cmake
└── libB
    ├── CMakeLists.txt
    └── CPM.cmake
```
**CMakeLists.txt**:
```cmake
cmake_minimum_required(VERSION 3.16.3)
project(minimal_bug_cpm)

add_subdirectory(libA)
add_subdirectory(libB)
```
**libA/CMakeLists.txt**:
```cmake
cmake_minimum_required(VERSION 3.16.3)
project(libA)

include(${CMAKE_CURRENT_LIST_DIR}/CPM.cmake)
CPMFindPackage(
  NAME doctest
  GITHUB_REPOSITORY onqtam/doctest
  GIT_TAG 2.4.4
)
```

**libB/CMakeLists.txt**:
```cmake
cmake_minimum_required(VERSION 3.16.3)
project(libB)

include(${CMAKE_CURRENT_LIST_DIR}/CPM.cmake)
CPMFindPackage(
  NAME fmt
  GITHUB_REPOSITORY fmtlib/fmt
  GIT_TAG 6.2.1
)
```

---

**ERROR**:  
```bash
-- Detecting CXX compile features - done
-- CPM: adding package doctest@2.4.4 (2.4.4)
--  adding package fmt@6.2.1 (6.2.1)
CMake Error: File /CMakeLists.cmake.in does not exist.
CMake Error at /usr/share/cmake-3.16/Modules/FetchContent.cmake:891 (configure_file):
  configure_file Problem configuring file
Call Stack (most recent call first):
  /usr/share/cmake-3.16/Modules/FetchContent.cmake:1006 (__FetchContent_directPopulate)
  /usr/share/cmake-3.16/Modules/FetchContent.cmake:1047 (FetchContent_Populate)
  libA/CPM.cmake:561 (FetchContent_MakeAvailable)
  libA/CPM.cmake:425 (cpm_fetch_package)
  libA/CPM.cmake:196 (CPMAddPackage)
  libB/CMakeLists.txt:6 (CPMFindPackage)


-- Configuring incomplete, errors occurred!
```

## Solution

`include(FetchContent)` need to be call inside each subdirectory (cmake < 3.17) else it's not working.
[...]

PR #193, opened 2021-02-06
fix superbuild cmake < 3.17

xmuller:
issue: #192 

* CPM_INDENT (first commit) :
I agree with you. set(CPM_INDENT "CPM:" CACHE INTERNAL "") seems to be appropriate.

* include(FetchContent):
Since the problem only occurs when using cmake < 3.17, I suggest to check the current cmake version and call `include(FetchContent)` for each `include(CPM.cmake)` when necessary.

During a future upgrade of the minimal cmake version, this can be noticed and removed easily.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -136,7 +136,10 @@ include(CMakeParseArguments)
 
 # Initialize logging prefix
 if(NOT CPM_INDENT)
-  set(CPM_INDENT "CPM:" CACHE INTERNAL "")
+  set(CPM_INDENT
+      "CPM:"
+      CACHE INTERNAL ""
+  )
 endif()
 
 function(cpm_find_package NAME VERSION)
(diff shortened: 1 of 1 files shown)


TheLartians:
Hey thanks for the PR! Yeah, the version fallback seems like a good approach. I wonder if we should also `include(CMakeParseArguments)`, as it's used by CPM.cmake functions as well.

xmuller:
For `include(CMakeParseArguments)` I didn't notice any problem. Since CMakeParseArguments only serves to manipulate variables I don't see how there would be an internal bug in cmake with this use case.

If I notice anything strange while continuing my project I will bring up the problem anyway.

TheLartians (approved):
Ah, I understood that the CMake include behaviour was causing the error, but if it's indeed due to an internal bug this approach makes perfect sense!

TheLartians:
Fix is live in v0.28.3! 🎉

PR #197, opened 2021-02-12
Update yaml-cpp git tag in README

dstansby:
`yaml-cpp` 0.6.3 is released, so I've updated the tag in the README to the tag for 0.6.3 (see https://github.com/jbeder/yaml-cpp/releases/tag/yaml-cpp-0.6.3)

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,8 +283,7 @@ CPMAddPackage(
   GITHUB_REPOSITORY jbeder/yaml-cpp
   # 0.6.2 uses deprecated CMake syntax
   VERSION 0.6.3
-  # 0.6.3 is not released yet, so use a recent commit
-  GIT_TAG 012269756149ae99745b6dafefd415843d7420bb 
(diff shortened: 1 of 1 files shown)


TheLartians:
Hey, thanks for the PR! We've [already updated](https://github.com/cpm-cmake/CPM.cmake/blob/master/examples/yaml/CMakeLists.txt) `yaml-cpp` (and other examples) in a previous [PR](https://github.com/cpm-cmake/CPM.cmake/pull/190).

TheLartians:
@dstansby I just now realised that you've updated the Readme while the PR handled the examples ... Sorry for not checking properly. You're definitely right that the example in the Readme could be updated as well! If you want, you can reopen the PR and I'll be happy to review and add the changes!

On that note, if you want to reopen, it would be great of you could reference the tag `yaml-cpp-0.6.3` instead of the commit hash, as it is more clear to be synchronised to the official release.

Issue #152, opened 2020-09-16
Allowing EXCLUDE_FROM_ALL


alandefreitas:
Hi @TheLartians ,

Thanks for the library.

When I include a CMake library with CPM.cmake it runs all the install commands in that library even when I don't need these installers with my own library. 

When adding the library as a subdirectory, it's possible to use `add_subdirectory(libname EXCLUDE_FROM_ALL)` to avoid that behavior.

Is it possible to use EXCLUDE_FROM_ALL when downloading a CMake library?

I don't know if that's a question or an enhancement.

Best,

TheLartians:
Hey, that's a good question. Currently I don't think there is a good way to achieve this behaviour besides setting the `DOWNLOAD_ONLY` flag and calling `add_subdirectory` manually. 

I actually wasn't really aware of this option until now, but it does sound like it could be the desired behaviour for most use-cases (given the option to opt-out). Will setting `EXCLUDE_FROM_ALL` on a directory still allow us to install all targets that are used normally or are there some caveats?

alandefreitas:
Setting `DOWNLOAD_ONLY` to `TRUE` and calling `add_subdirectory` manually is something I didn't think about. This should be able to achieve something more similar to `find_package`. It's a great option. What a Columbus' egg. It's not a replacement for `find_package` in the sense that I'll need some conditionals after that but it's more than enough for what I need. Feel free to close this issue if you want. 

If you're still interested in the idea anyway, using `EXCLUDE_FROM_ALL` could be a good practice when using `add_subdirectory` as a way to emulate what `find_package` does. I think that's why they created this option. Because when you use `find_package` it doesn't mess with your `install` commands and these imported targets don't get built when you build `all`. Without `EXCLUDE_FROM_ALL`, `add_subdirectory` would never be able to replace `find_package` properly in any situation because `find_package` never puts its own targets in your `install` commands. Some libraries include an option to not build the install target for that reason, but not many libraries do that.

Regarding your question, as far as I understand, `EXCLUDE_FROM_ALL` does not forbid us to install these imported targets in any way. It just stops these scripts in the subdirectory from messing with the `install` commands in the main project by default with no way to opt-out. Putting external packages in your own package becomes an option rather than a requirement for all external packages. By default, it behaves a little as if you were using `find_package`. It's also great not to have lots of useless targets in your "build all" when you just want to use the subdirectory as if it were a package. I'm sorry if I didn't understand your question properly.

[...]

PR #198, opened 2021-02-15
Add dependencies using EXCLUDE_FROM_ALL flag

TheLartians:
Implements #152 by adding all dependencies ~using the `EXCLUDE_FROM_ALL` flag by default~ adding an `EXCLUDE_FROM_ALL` optional argument. ~Further testing is required to see this could represent a breaking change for real projects and see if an optional implementation or an opt-out flag would be appropriate.~

## TODO

- [x] Test in real-world projects
- [x] Decide if should be always on, opt-in or opt-out
- [x] Settle for an API
- [x] Update `cmake-format` configuration
- [x] Consider edge cases
- [x] Add a unit test


TheLartians:
After some testing I believe that the option should not be set by default, as even some of my own projects depend on installation of components defined only in outer dependencies. This might not be a best practice, but setting the flag by default could unintentionally break other projects as well. A smoother transition may be achieved by adding an optional flag that may be promoted to default sometime in the future.

PR #202, opened 2021-02-17
If a name isn't provided, try to infer it from the git repo

iboB:
Two reasons behind this PR

**More compact CPMAddPackage calls**

```cmake
CPMAddPackage(
  GITHUB_REPOSITORY catchorg/Catch2
  VERSION 2.5.0
)
```

**Less room for error in pasting CPMAddPackage calls**

To me, and I suspect to many other users most packages look like this:

```cmake
CPMAddPackage(
  NAME PackageName
  GITHUB_REPOSITORY user/PackageName
  VERSION x.y.z
)
```

When adding a new package, this gets pasted around and two identical occurrences of "PackageName" need to be replaced. If we follow the DRY principle, this repetition should be avoided and in the case of git repositories this is easy.

The old behavior still works for backwards compatibility and for the rare cases where the package name and the repo name aren't the same.


cmake/CPM.cmake
```diff
@@ -134,6 +134,17 @@ endif()
...
 
+# Infer package name from git repository uri (path or url)
+function(cpm_package_name_from_git_uri URI RESULT)
+  string(REGEX MATCH "([^/:]+)/?.git/?$" cpmGitUriMatch "${URI}")
+  if(DEFINED cpmGitUriMatch)
+    set(${RESULT}
+        ${CMAKE_MATCH_1}
+        PARENT_SCOPE
+    )
+  endif()
```
TheLartians, line 145: It may be a good idea to unset `${RESULT}` in the parent scope when no match is found. Otherwise the value could be mistaken with an unrelated outside variable using the same name.

TheLartians (commented):
Thanks for the PR, I think this is a great idea to shorten the syntax and avoid repetition!

TheLartians:
Also feel free to update the main example in the readme to show off the shorter syntax! :-)

cmake/CPM.cmake
```diff
@@ -309,6 +319,12 @@ function(CPMAddPackage)
     endif()
   endif()
 
+  # Check for required arguments
+
+  if(NOT DEFINED CPM_ARGS_NAME)
+    message(FATAL_ERROR "CPM: 'NAME' was not provided for package added with arguments: '${ARGN}'")
```
TheLartians, line 325: I see that this will become the predominant way of adding packages. Perhaps it may be good to explain here that `NAME` could not be inferred automatically, as it may confuse new users.

```suggestion
    message(FATAL_ERROR "CPM: 'NAME' was not provided and couldn't be automatically inferred for package added with arguments: '${ARGN}'")
```

iboB:
Check. Suggested changes were added.

TheLartians (approved):
Awesome, thanks for the changes!

TheLartians:
Feature released in v0.30.0! 🎉

GunZi200:
@TheLartians 
Hi, I'm glad to see all the improvements... I have one idea to improve the README:

This is what it looks like right now:

> ExternalProject works similarly as FetchContent, however waits with adding dependencies until build time. This has a quite a few disadvantages, especially as it makes using custom toolchains / cross-compiling very difficult and can lead to problems with nested dependencies.

I think it makes sense to highlight the keywords ExternalProject and FetchContent like this: `ExternalProject` `FetchContent` or even link the words to the CMake documentation.

## Idea 1
> `ExternalProject` works similarly as `FetchContent`, however waits with adding dependencies until build time. This has a quite a few disadvantages, especially as it makes using custom toolchains / cross-compiling very difficult and can lead to problems with nested dependencies.

----

I just learned right now as I wrote this that you can create a link for coded words with 
* \`keyword \` (`keyword`) :D)

Or like this: 

## Idea 2

> [`ExternalProject`](https://cmake.org/cmake/help/latest/module/ExternalProject.html) works similarly as [`FetchContent`](https://cmake.org/cmake/help/latest/module/FetchContent.html) however waits with adding dependencies until build time. This has a quite a few disadvantages, especially as it makes using custom toolchains / cross-compiling very difficult and can lead to problems with nested dependencies.

TheLartians:
@GunZi200 thanks for the suggestions! I think there are definitely some improvements that can be made to the readme. As the section you're referring to is written for users who already know both commands I think the link isn't necessary but highlighting the code does make sense. If you want, feel free to open a PR and we can discuss further changes there.

PR #206, opened 2021-02-18
Improved regex match in cpm_package_name_from_git_uri

iboB:
The old code has an issue (though not a bug). 

Apparently `cpmGitUriMatch` always gets defined, so we never end up in the `else`. However setting a variable to an empty value (`CMAKE_MATCH_1`) is equivalent to unset and that's why the tests pass and there is no bug.

The new code is cleaner and doesn't rely on an unused temporary.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -136,8 +136,7 @@ include(CMakeParseArguments)
 
 # Infer package name from git repository uri (path or url)
 function(cpm_package_name_from_git_uri URI RESULT)
-  string(REGEX MATCH "([^/:]+)/?.git/?$" cpmGitUriMatch "${URI}")
-  if(DEFINED cpmGitUriMatch)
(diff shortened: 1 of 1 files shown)


TheLartians (approved):
Interesting, so in case there was no match, I guess `${cpmGitUriMatch}` would be defined to an empty string? But I do agree that matching without the temporary seems cleaner.

On a side note this is the first time I've seen an `if` clause have side effects (defining the `CMAKE_MATCH_X` variables). 😄

iboB:
> Interesting, so in case there was no match, I guess `${cpmGitUriMatch}` would be defined to an empty string? 

Yes, exactly. It was a surprise to me too.

```cmake
set(foo "") # sets foo to empty string, foo is DEFINED
set(foo) # sets foo to empty value, equivalent to unset
```

Apparently `string(REGEX MATCH` always defines the output variable. If there is no match, it's defined to an empty string.

PR #218, opened 2021-02-22
Update examples to use shorthand syntax where possible

TheLartians:
To further propagate the new shorthand syntax for suited CPM.cmake dependencies, this PR updates the examples to use the new Syntax where possible. Also consistently adds quotation marks, potentially avoiding confusion for the `#` syntax, as is would be interpreted as a comment without quotes. Also clarifies the `EXCLUDE_FROM_ALL` behaviour in the readme.

### DIFF
--- a/examples/doctest/CMakeLists.txt
+++ b/examples/doctest/CMakeLists.txt
@@ -6,17 +6,8 @@ project(CPMExampleDoctest)
 
 include(../../cmake/CPM.cmake)
 
-CPMAddPackage(
-  NAME fibonacci
-  GITLAB_REPOSITORY TheLartians/Fibonacci
-  VERSION 2.0
-)
-
-CPMAddPackage(
-  NAME doctest
-  GITHUB_REPOSITORY onqtam/doctest
-  GIT_TAG 2.4.5
-)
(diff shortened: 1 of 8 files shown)


examples/doctest/CMakeLists.txt
```diff
@@ -6,17 +6,8 @@ project(CPMExampleDoctest)
...
-  GITLAB_REPOSITORY TheLartians/Fibonacci
-  VERSION 2.0
-)
-
-CPMAddPackage(
-  NAME doctest
-  GITHUB_REPOSITORY onqtam/doctest
-  GIT_TAG 2.4.5
-)
+CPMAddPackage("gh:TheLartians/Fibonacci@2.0")
```
iboB, line 9: note that this changes the name of the package from `fibonacci` to `Fibonacci` - capitalized

iboB (approved):

examples/doctest/CMakeLists.txt
TheLartians: Good point it, imo the repo name `Fibonacci` should have been the correct name from the start. Shows that the new single-arg syntax is less error prone thanks to avoided repetition! 👍

PR #220, opened 2021-02-23
Infer package name and version from URL

### LINKED ISSUE #208: Infer package name from URL
#181 is about inferring the version from a URL. It should be extended to also infer the name if not provided (much like #202 introduced inferring a name from a git repository).

If we can infer the name from a URL, fixing several lines in #207 (as indicated in the code), will allow URL-based packages to be added by the shorthand syntax .

@jhasse doesn't seem to be responding, so I volunteer to be assigned to this and move forward both version and name inference from URLs.

iboB:
This is both for the single-argument syntax, and for the no-name-multi-argument syntax

Closes #208 

Closes #181 unmerged. It's superseded by this


cmake/CPM.cmake
```diff
@@ -397,6 +438,24 @@ function(CPMAddPackage)
     endif()
   endif()
 
+  if(DEFINED CPM_ARGS_URL)
+    # If a name or version aren't provided, try to infer them from the git repo
```
TheLartians, line 442:
```suggestion
    # If a name or version aren't provided, try to infer them from the URL
```

cmake/CPM.cmake
```diff
@@ -397,6 +438,24 @@ function(CPMAddPackage)
...
+    list(GET CPM_ARGS_URL 0 firstUrl)
+    cpm_package_name_and_ver_from_url(${firstUrl} nameFromUrl verFromUrl)
+    # If we fail to obtain name and version from the first URL, we could try other URLs if any.
+    # However multiple URLs are expected to be quite rare, so for now we won't bother.
```
TheLartians, line 446: TIL `FetchContent` supports multiple URLs 😄

README.md
```diff
@@ -62,6 +62,17 @@ CPMAddPackage("uri@version#tag")
 
 In the shorthand syntax if the URI is of the form `gh:user/name`, it is interpreted as GitHub URI and converted to `https://github.com/user/name.git`. If the URI is of the form `gl:user/name`, it is interpreted as a [GitLab](https://gitlab.com/explore/) URI and coverted to `https://gitlab.com/user/name.git`. Otherwise the URI used verbatim as a git URL. All packages added using the shorthand syntax will be added using the [EXCLUDE_FROM_ALL](https://cmake.org/cmake/help/latest/prop_tgt/EXCLUDE_FROM_ALL.html) flag.
 
+The single-argument syntax also works for URLs:
+
+```cmake
+# An archive package from a given url. The version is inferred
+CPMAddPackage("https://example.com/my-package-1.2.3.zip")
+# An archive package from a given url with an MD5 hash provided
+CPMAddPackage("https://example.com/my-package-1.2.3.zip#68e20f674a48be38d60e129f600faf7d")
```
TheLartians, line 71: Doesn't CMake use `URL_HASH`s in the form [`URL_HASH <algo>=<hashValue>`](https://cmake.org/cmake/help/latest/module/ExternalProject.html)? So should the syntax rather be like

```cmake
CPMAddPackage(
  "https://example.com/my-package-1.2.3.zip#MD5=68e20f674a48be38d60e129f600faf7d"
)
```

Or is MD5 the default algorithm? I think we should probably document and test either case.

TheLartians (commented):
Thanks for the PR, definitely awesome to support the shorthand syntax for URLs as well!

cmake/CPM.cmake
iboB: I did recently too: https://gitlab.kitware.com/cmake/cmake/-/issues/21847

README.md
iboB: MD5 is the default. But I'll document that others are possible

TheLartians: Could we also get a test case for `<algo>=<hash>`? Just so future changes in the regex won't break the feature.

iboB: Yes. Will do in a couple of hours.
BTW even though I clearly remember reading that MD5 is the default, I can't find this anywhere. I opened a [issue on CMake's tracker](https://gitlab.kitware.com/cmake/cmake/-/issues/21859) to make sure

iboB:
OK. I added an explicit algorithm in README.md. We can document a default once it's confirmed in the CMake issue.

TheLartians (approved):
Awesome, thanks!

TheLartians:
Feature is live in v0.31.1 🎉

PR #223, opened 2021-02-25
update more examples to use shorthand syntax

TheLartians:
Thanks to `EXCLUDE_FROM_ALL` being active by default, we should be able to update even more examples to use the awesome new shorthand syntax without worrying about leaking examples or tests into the main build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,44 +82,6 @@ After calling `CPMAddPackage` or `CPMFindPackage`, the following variables are d
 The difference between `CPMFindPackage` and `CPMAddPackage` is that `CPMFindPackage` will try to find a local dependency via CMake's `find_package` and fallback to `CPMAddPackage` if the dependency is not found.
 This behaviour can be also modified globally via [CPM options](#options).
 
-## Full CMakeLists Example
-
-```cmake
-cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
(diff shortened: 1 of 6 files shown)


TheLartians:
Sorry for spamming with the commits @iboB, seems it's a bit too early for me to do pull requests 😄

iboB:
Just to make sure: With `EXCLUDE_FROM_ALL YES`, these examples and tests which are now not explicitly disabled won't show up in the build, because they are not referenced from the project which is being configured. Is this right?

TheLartians:
[EXCLUDE_FROM_ALL](https://cmake.org/cmake/help/latest/prop_tgt/EXCLUDE_FROM_ALL.html) essentially prevents any inner targets from being added to the `all` target. So while still part of the build, they are now only built if the user explicitly calls `cmake --build build --target <excluded-target-name>`. Something I haven't checked is if the dependency's tests get added to the `tests` target, but if so it should be filed and reported as a CMake bug imo.

iboB (approved):

TheLartians:
Turns out, `EXCLUDE_FROM_ALL` disables the targets, but still allows the tests. 
A relevant CMake issue already exists: https://gitlab.kitware.com/cmake/cmake/-/issues/20212.

PR #235, opened 2021-03-22
Scope package options to avoid changing the local scope

### LINKED ISSUE #222: Parent project options are being cancelled
If the parent project has an option that conflicts with the child project, the child project is messing with the scope of the parent project.

For instance

```cmake
message("BUILD_EXAMPLES=${BUILD_EXAMPLES}")
CPMAddPackage(
        NAME pareto
        GIT_REPOSITORY https://github.com/alandefreitas/pareto.git
        GIT_TAG v1.2.0
        OPTIONS
        "BUILD_MATPLOT_TARGETS OFF"
        "BUILD_EXAMPLES OFF"
)
add_executable(spatial_containers spatial_containers.cpp)
target_link_libraries(spatial_containers PUBLIC pareto)
message("BUILD_EXAMPLES=${BUILD_EXAMPLES}")
```

gives me 

```cmake
BUILD_EXAMPLES=ON
-- CPM: adding package pareto@1.2.0 (v1.2.0)
-- Looking for min
-- Looking for min - not found
BUILD_EXAMPLES=OFF
```

which is a problem because some option names are quite common. Some projects create have prefixes in their options but most projects don't.

Of course, people can manually do 

```cmake
set(BUILD_EXAMPLES_PREV ${BUILD_EXAMPLES})
message("BUILD_EXAMPLES=${BUILD_EXAMPLES}")
CPMAddPackage(
        NAME pareto
        GIT_REPOSITORY https://github.com/alandefreitas/pareto.git
        GIT_TAG v1.2.0
        OPTIONS
        "BUILD_EXAMPLES OFF"
)
add_executable(spatial_containers spatial_containers.cpp)
target_link_libraries(spatial_containers PUBLIC pareto)
set(BUILD_EXAMPLES ${BUILD_EXAMPLES_PREV} CACHE BOOL "" FORCE)
message("BUILD_EXAMPLES=${BUILD_EXAMPLES}")
```

for all options but that defeats the purpose of the OPTIONS arguments, which is to do what we cannot already do with FetchContent.

flagarde:
This is the fix for #222, but I found a strange behaviour. Bellow is a test code :
```cmake
cmake_minimum_required(VERSION 3.14)
project(TestOptions)

include("cmake/CPM.cmake")

set(SET "main")
message(STATUS "From main before set:${SET}, option: ${OPT}")


CPMAddPackage(
  NAME titi
  GIT_REPOSITORY https://gitlab.com/flagarde/titi
  GIT_TAG main 
  OPTIONS "OPT ON" "SET YU"
)
message(STATUS "From main between set:${SET}, option: ${OPT}")

CPMAddPackage(
  NAME toto
  GIT_REPOSITORY https://gitlab.com/flagarde/toto
  GIT_TAG main 
  OPTIONS "OPT OFF" "SET CHANGE"
)

message(STATUS "From main after set:${SET}, option: ${OPT}")
```

First run gives:
```bash
-- The C compiler identification is GNU 10.2.0
-- The CXX compiler identification is GNU 10.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- From main before set:main, option: 
-- CPM: adding package titi@ (main)
-- From titi, set: titi default was "titi" , option: ON default was "FALSE"
-- From main between set:main, option: 
-- CPM: adding package toto@ (main)
-- From toto, set: CHANGE default was "toto" , option: OFF default was "TRUE"
-- From main after set:main, option: 
-- Configuring done
-- Generating done
```
`-- From titi, set: titi default was "titi" , option: ON default was "FALSE" ` should be `-- From titi, set: YU default was "titi" , option: ON default was "FALSE" `.

Second gives the good results : 
```bash
-- From main before set:main, option: 
-- CPM: adding package titi@ (main)
-- From titi, set: YU default was "titi" , option: ON default was "FALSE"
-- From main between set:main, option: 
-- CPM: adding package toto@ (main)
-- From toto, set: CHANGE default was "toto" , option: OFF default was "TRUE"
-- From main after set:main, option: 
-- Configuring done
-- Generating done
```


TheLartians:
Thanks for the PR and fix! I believe that `cpm_fetch_package` is not the optimal location to add options though, as we circumvent `FetchContent` if a dependency is cached locally. I've updated the PR to set the options in `cpm_add_subdirectory` instead. We should probably write a test as well.

TheLartians:
Interestingly, on Travis CI the cereal example now fails with 

```
running example cereal
1128======================
1129- cmake -Hexamples/cereal -Bbuild/cereal
1130  -- CPM: adding package cereal@1.3.0 (v1.3.0)
1131- cmake --build build/cereal -- -j1
1132error while running 'cmake --build build/cereal -- -j1':
1133   In file included from /home/travis/build/cpm-cmake/CPM.cmake/build/cereal/_deps/cereal-src/./include/cereal/cereal.hpp:32,
1134                   from /home/travis/build/cpm-cmake/CPM.cmake/build/cereal/_deps/cereal-src/./include/cereal/types/memory.hpp:33,
1135                   from /home/travis/build/cpm-cmake/CPM.cmake/build/cereal/_deps/cereal-src/unittests/common.hpp:30,
1136                   from /home/travis/build/cpm-cmake/CPM.cmake/build/cereal/_deps/cereal-src/unittests/memory.hpp:29,
1137                   from /home/travis/build/cpm-cmake/CPM.cmake/build/cereal/_deps/cereal-src/unittests/memory.cpp:28:
1138  /usr/include/c++/8/type_traits:38:10: fatal error: bits/c++config.h: No such file or directory
1139   #include <bits/c++config.h>
1140            ^~~~~~~~~~~~~~~~~~
1141  compilation terminated.
1142  make[2]: *** [_deps/cereal-build/unittests/CMakeFiles/test_memory_32.dir/memory.cpp.o] Error 1
1143  make[1]: *** [_deps/cereal-build/unittests/CMakeFiles/test_memory_32.dir/all] Error 2
1144  make: *** [all] Error 2
1145
```

Either something changed in their runtime or our changes are somehow responsible for the error.

TheLartians:
Turns out that CMake policies get reset when `cmake_minimum_required` is called. To fix this, [we need to set `CMAKE_POLICY_DEFAULT_CMP0077`as well](https://gitlab.kitware.com/cmake/cmake/-/issues/20312).

TheLartians:
Note that this may be a breaking change in case users relied on the variable caching or old behaviour of the CPM007 policy.

TheLartians:
Another counter-intuitive behaviour is that even though the option is overridden in the local scope, CMake still defines and sets the option to its default value in the cache. It may make sense to [mark the specified options as advanced](https://cmake.org/cmake/help/latest/command/mark_as_advanced.html), as their behaviour cannot be overwritten from the CMake UI.

TheLartians:
@flagarde sorry for taking over the PR! I expected this to be a small change to fix the CI, but it kinda escalated thanks to CMake weirdness. I'll leave it now. 😅

iboB:
@TheLartians not sure if you still need my review for this, but I won't have time to properly check it out in the following 2-3 days

TheLartians:
@iboB no problem, we can let this sit a few more days! I would definitely appreciate a contributor review and perhaps also another opinion on this topic as it has potential of breaking somebody's workflow.

flagarde:
> Another counter-intuitive behaviour is that even though the option is overridden in the local scope, CMake still defines and sets the option to its default value in the cache. It may make sense to [mark the specified options as advanced](https://cmake.org/cmake/help/latest/command/mark_as_advanced.html), as their behaviour cannot be overwritten from the CMake UI.

then we could maybe explicitly `unset` it should be quite easy and maybe safer?

flagarde:
Yes it looks like, but the way it will be done would allow better encapsulation I think. I don't know if the behaviour you describe it's the expected behaviour by CMake developers or if we are pointing out a corner case?

flagarde:
As you pointed out the policy 0077 could crash some project but you have the CMake rules from our side so it would be easy to convince people to "correct" their repo. One day or an other this policy will be activated

cmake/CPM.cmake
```diff
@@ -132,7 +132,6 @@ if(NOT CPM_DONT_CREATE_PACKAGE_LOCK)
 endif()
 
 include(FetchContent)
-include(CMakeParseArguments)
```
iboB, line 135: why is this line removed?

iboB (changes requested):
I think we should also try to test the bug that the issue is supposed to fix.

Have an ON option in the root test project and set it to OFF in the dependency project, then check that it's not reset in the root

iboB:
BTW @TheLartians, how did you make commits in @flagarde's fork?

johelegp:
> BTW @TheLartians, how did you make commits in @flagarde's fork?

See https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests.

iboB:
As a whole I tend to agree that the issue needs to be addressed. CPM is not popular enough (yet) in order to make demands from library authors. If it was, I would be in favor of the "Just scope your options!" solution, but sadly we have to live with dependencies which have clashing values.

> Another counter-intuitive behaviour is that even though the option is overridden in the local scope, CMake still defines and sets the option to its default value in the cache.

I'm not sure I understand this. Is this about the dependency's option which is overridden by `CPMAddPackage`'s `OPTIONS`? The value from the dependency get cached? What happens if the parent project has the same option? What gets cached then? 

As a whole I don't know what CMake does when projects and subdirs have clashing options. Where can I read about it.

My general concerns which are not related to the code but to the solution are:
Assuming the dependency has this line `option(BUILD_TESTS "Dep: building tests" OFF)`

* What happens if the parent sets an **option**  `BUILD_TESTS` to `ON`?
* What happens if the parent sets an uncached **variable** `BUILD_TESTS` to `ON`?
* What if it's cached?
* If the parent calls `CPMAddPackage(Dep OPTIONS "BUILD_TESTS ON")` but otherwise doesn't touch that symbol in its code is `BUILD_TESTS` still present in CMake GUI-based cofigs?
* If yes, what happens if we configure the parent with `-DBUILD_TESTS=0`?

flagarde:
> I tend to agree that the issue needs to be addressed. CPM is not popular enough (yet) in order to make demands from library authors. If it was, I would be in favor of the "Just scope your options!" solution, but sadly we have to live with dependencies which have clashing values.
> 
> > Another counter-intuitive behaviour is that even though the option is overridden in the local scope, CMake still defines and sets the option to its default value in the cache.
> 
> I'm not sure I understand this. Is this about the dependency's option which is overridden by `CPMAddPackage`'s `OPTIONS`? The value from the dependency get cached? What happens if the parent project has the same option? What gets cached then?
> 
> Unfortunately I simply don't know what CMake does when projects and subdirs have clashing options. Where can I read about it?
> 
> My general concerns which are not related to the code but to the solution are:
> 
> _Assuming the dependency has this line `option(BUILD_TESTS "Dep: building tests" OFF)`_
> 
>     * What happens if the parent sets an **option**  `BUILD_TESTS` to `ON`?
> 
>     * What happens if the parent sets an uncached **variable** `BUILD_TESTS` to `ON`?
> 
>     * What if it's cached?
> 
>     * If the parent calls `CPMAddPackage(Dep OPTIONS "BUILD_TESTS ON")` but otherwise doesn't touch that symbol in its code is `BUILD_TESTS` still present in CMake GUI-based cofigs?
> 
>     * If yes, what happens if we configure the parent with `-DBUILD_TESTS=0`?

Are you talking about what the code is doing or what it is supposed to do ?
From my understanding  

with policy 0077 the option is ignored and the  parent give his value to childs. If you want to set child off pass it as option explicitly in `CPMAddPackage`. For me it makes sense that the parent take precedence over childs right? (https://cmake.org/cmake/help/latest/command/option.html#command:option)

So in principle option cached or normal variable should have precedence over the childs ones.

This behaviour it's the one I would expect but maybe i'm wrong. If i really want to force BUILD_TESTS=OFF in child I would put it in OPTIONS when i call  `CPMAddPackage`. Please correct me if this behaviour looks strange for you, or if it's not what you want CPM to do.

cmake/CPM.cmake
TheLartians: Since CMake 3.4 the module is [implemented natively](https://cmake.org/cmake/help/latest/module/CMakeParseArguments.html). As our minimum version is 3.14, we don't need the import.

TheLartians:
> I think we should also try to test the bug that the issue is supposed to fix.
> Have an ON option in the root test project and set it to OFF in the dependency project, then check that it's not reset in the root

We currently have a test that verifies that the variable overwrite is indeed scoped. With the assumption that CMake's scoping rules are consistent, this should already cover the bug's case.

Tbh I'm not that sure anymore about removing not previously defined options and cache variables, as this has the potential of introducing a new family of unexpected edge cases (e.g. the behaviour changes if a variable has been defined before or not), that can be just as bad as the issue we're trying to solve. I'd currently vote for simply [marking the options as advanced](https://cmake.org/cmake/help/latest/command/mark_as_advanced.html), indicating to users that changing the value could not have the consequences they might expect.

iboB:
Unfortunately I still don't follow. 

Which options are you suggesting to be marked as advanced? 

What is this removing of not previously defined options?

> Another counter-intuitive behaviour is that even though the option is overridden in the local scope, CMake still defines and sets the option to its default value in the cache.

What gets cached here?

TheLartians:
> What gets cached here?

Say we have a package foo with the following CMakeLists:

```cmake
option(IS_SUBDIRECTORY "this should be set to YES for external projects" NO)
[...]
```

And we add the package through CPM.cmake, then CMake keeps the default value, as we didn't overwrite the cache. This may be very unexpected to users.

```cmake
# `IS_SUBDIRECTORY ` is undefined here

CPMAddPackage(
  NAME foo
  [...]
  OPTIONS "IS_SUBDIRECTORY YES"
  # inside foo's CMakeLists, `IS_SUBDIRECTORY ` is `YES`
 )

# `IS_SUBDIRECTORY ` is `NO` here
```

> Which options are you suggesting to be marked as advanced?
> What is this removing of not previously defined options?

These are potential actions we could take to address the problem. 
- By removing the variables after setting the options we can avoid options leaking into the outer scope with unexpected values.
- We could also reset them to their previous state, thus making them immutable in the parent (even through changing the cache) by the inner dependency.
- We could mark them as advanced, indicating to users that they are not meant to be changed by the CMake UI, or that there are caveats.

flagarde:
It's a strange behavior. Does policy 0077 https://cmake.org/cmake/help/latest/policy/CMP0077.html wouldn't apply here in the child repo ?

>> The NEW behavior for this policy is to do nothing when a normal variable of the same name exists. The normal variable is not removed. The cache entry is not created or updated and is ignored if it exists.

I think that 
```cmake
# allows us to change options with uncached variables
    cmake_policy(SET CMP0077 NEW)
    set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
```
is inside the function `cpm_add_subdirectory` so it is not seen when you apply the fetch. Could you try to put this in CPMAddPackage

TheLartians:
Ok I might just be dumb and am unable to reproduce the error. Maybe I forgot to clean some caches during testing. 🙈 
In that case the behaviour may just be exactly as expected!

I think the position inside `cpm_add_subdirectory` is correct though, as that's the only place where we actually add subdirectories. (we don't use `FetchContent_makeavailable` anymore)

flagarde:
I thought CMake was making its magics under FetchContent_Populate which is a wrapper to FetchContent_Populate( <name> ) wich have a configure step.

iboB:
> Ok I might just be dumb and am unable to reproduce the behaviour. Maybe I forgot to clean some caches during testing. 🙈
> In that case the behaviour may just be exactly as expected!

🤣  

Ok... to recap. The behavior with this PR merged is the following:

```cmake
# Dependency
option(BUILD_TESTS "dep: build tests" OFF)
option(BUILD_EXAMPLES "dep: build tests" OFF)
option(SOMETHING_ELSE "dep: build tests" OFF)
...
```

```cmake
# root
option(BUILD_EXAMPLES "root: build texamples" ON)
set(SOMETHING_ELSE ON)
CPMAddPackage(Dependency 
  OPTIONS "BUILD_TESTS ON"
)

# here 
# BUILD_EXAMPLES is ON
# Dependency was configured with BUILD_EXAMPLES ON
# BUILD_EXAMPLES is in the cache as "root: build texamples" ON
# BUILD_TESTS is not defined
# Dependency was configured with BUILD_TESTS ON
# BUILD_TESTS is in the cache as ON "dep: build tests"
# SOMETHING_ELSE is ON
# Dependency was configured with SOMETHING_ELSE ON
# SOMETHING_ELSE is not in the cache
```

If this is true, then I approve of this PR `[[iboB seal of approval.png]]`

flagarde:
> > Ok I might just be dumb and am unable to reproduce the behaviour. Maybe I forgot to clean some caches during testing. see_no_evil
> > In that case the behaviour may just be exactly as expected!
> 
> rofl
> 
> Ok... to recap. The behavior with this PR merged is the following:
> 
> ```cmake
> # Dependency
> option(BUILD_TESTS "dep: build tests" OFF)
> option(BUILD_EXAMPLES "dep: build tests" OFF)
> option(SOMETHING_ELSE "dep: build tests" OFF)
> ...
> ```
> 
> ```cmake
> # root
> option(BUILD_EXAMPLES "root: build texamples" ON)
> set(SOMETHING_ELSE ON)
> CPMAddPackage(Dependency 
>   OPTIONS "BUILD_TESTS ON"
> )
> 
> # here 
> # BUILD_EXAMPLES is ON
> # Dependency was configured with BUILD_EXAMPLES ON
> # BUILD_EXAMPLES is in the cache as "root: build texamples" ON
> # BUILD_TESTS is not defined
> # Dependency was configured with BUILD_TESTS ON
> # BUILD_TESTS is NOT in the cache
> # SOMETHING_ELSE is ON
> # Dependency was configured with SOMETHING_ELSE ON
> # SOMETHING_ELSE is NOT in the cache
> ```
> 
> If this is true, then I approve of this PR `[[iboB seal of approval.png]]`

@iboB It looks fine if 'not defined' means OFF, I didn't know OFF means not defined for CMake

iboB:
@flagarde I mean not defined, not OFF. OFF is still defined..,

So the aforementioned behavior that the overridden options leak their default initialization values to the root still remains.

I think we can live with that. As @TheLartians said, this is a CMake-specific issue and not a CPM issue.

iboB (approved):

TheLartians:
@iboB I've updated the test to check options inspired by your example. I currently believe the behaviour is as you said!
@flagarde sorry for taking over again, the merge with master looks a bit complex as well, give me a sec 😄

flagarde:
@TheLartians Maybe it would be nice to explain the `aforementioned behavior that the overridden options leak their default initialization values to the root still remains.` in the doc as it could have strange effect even if it seems not to be a CPM problem but CMake one.

iboB (approved):

TheLartians:
@flagarde actually I believe that was my mistake due to not properly cleaning my caches as I can't reproduce the behaviour. You can see in the new test case that it appears that overwritten options aren't leaked after all, which would be the expected behaviour.

PR #243, opened 2021-03-26
Check options

flagarde:
Hi,

Restore the ability to check OPTIONS.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -246,7 +246,7 @@ function(CPMFindPackage)
     return()
   endif()
 
-  cpm_check_if_package_already_added(${CPM_ARGS_NAME} "${CPM_ARGS_VERSION}")
+  cpm_check_if_package_already_added(${CPM_ARGS_NAME} "${CPM_ARGS_VERSION}" "${CPM_ARGS_OPTIONS}")
   if(CPM_PACKAGE_ALREADY_ADDED)
     cpm_export_variables(${CPM_ARGS_NAME})
     return()
@@ -262,7 +262,7 @@ function(CPMFindPackage)
 endfunction()
 
 # checks if a package has been added before
-function(cpm_check_if_package_already_added CPM_ARGS_NAME CPM_ARGS_VERSION)
+function(cpm_check_if_package_already_added CPM_ARGS_NAME CPM_ARGS_VERSION CPM_ARGS_OPTIONS)
   if("${CPM_ARGS_NAME}" IN_LIST CPM_PACKAGES)
     CPMGetPackageVersion(${CPM_ARGS_NAME} CPM_PACKAGE_VERSION)
     if("${CPM_PACKAGE_VERSION}" VERSION_LESS "${CPM_ARGS_VERSION}")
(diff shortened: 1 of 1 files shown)


cmake/CPM.cmake
```diff
@@ -773,10 +834,32 @@ function(cpm_parse_option OPTION)
       "${OPTION_KEY}"
       PARENT_SCOPE
   )
-  set(OPTION_VALUE
-      "${OPTION_VALUE}"
-      PARENT_SCOPE
+  # https://cmake.org/cmake/help/latest/command/if.html#basic-expressions (1/0 is problematic)
```
iboB, line 837: why is 1/0 problematic?

cmake/CPM.cmake
```diff
@@ -271,6 +271,55 @@ function(cpm_check_if_package_already_added CPM_ARGS_NAME CPM_ARGS_VERSION)
           "${CPM_INDENT} requires a newer version of ${CPM_ARGS_NAME} (${CPM_ARGS_VERSION}) than currently included (${CPM_PACKAGE_VERSION})."
       )
     endif()
+
+    if(CPM_ARGS_OPTIONS)
```
iboB, line 275: What if when the that package was first added options were not set and not they are, or vice-versa?

cmake/CPM.cmake
flagarde: It could mean `"ON"` or `NUMBER_OF_CORE "1"`

cmake/CPM.cmake
flagarde: Yes maybe this cases are not well checked.

cmake/CPM.cmake
iboB: Then why change all values to "1"? Instead of to "ON". Thus this case will be covered

TheLartians (commented):
Hey @flagarde, thanks for re-implementing the options check!

Thinking a bit more about the options, I'm not sure if it's a good idea to perform the check in all cases. E.g. often an upstream package would override a downstream version with a newer dependency. In that case the option names and behaviours could have changed, so the old options no longer apply. 

Perhaps we should only perform the check if the versions match, or at least offer a way to opt-out?

cmake/CPM.cmake
flagarde: @TheLartians yes good point, this it's an other problem. Maybe CPM could accept only TRUE YES ON. It makes it more informative and I'm  temptated to say using 1 or Y for TRUE should be avoided.  `I think we should either use the normalisation only when checking of parameter consistency or add a note in the readme that we recommend truthy/falsy values to be standardised in CPM.cmake options (e.g. to YES/NO).` I agree totlaly to this point but maybe we could add TRUE and On as they are not preblematic and very well used. I have never seen 1 or Y.

flagarde:
> Hey @flagarde, thanks for re-implementing the options check!
> 
> Thinking a bit more about the options, I'm not sure if it's a good idea to perform the check in all cases. E.g. often an upstream package would override a downstream version with a newer dependency. In that case the option names and behaviours could have changed, so the old options no longer apply.
> 
> Perhaps we should only perform the check if the versions match, or at least offer a way to opt-out?

If I understand how CPM is working now, if the upstream ask for a newer version a warning will be raised and the new version will be used. I think this are two different problems and it would be nice to have the warning for the version and the warning for the options. 

If you are not warned you could disable some option necessary downstream without knowing it. Sometimes you don't know exactly what are the need of the used repositories. It could be possible to add an option to disable the check. 

Internally CMake use ON so if you want to force only one key maybe you should choose this one

flagarde:
In case you include two time the same repo with the same version : Do you want to keep the previous option or the last one. I suppose you want to keep the last one ?

In cas you have same repo but two version you want new version and the new options right ?

TheLartians:
Well with the way CPM.cmake currently works (we immediately add the package after definition), we always use the first version and options added. In case another dependency asks for a newer version that the one currently added, we inform the user it may be necessary to upgrade by outputting a warning. As for options I think we must ask ourselves which scenarios are likely. E.g. if a project adds the same dependency twice, but using different options, I see the following scenarios:

- The version is the same as before: there is a chance that the user was unaware of the other requirement and the options of both should be merged. Other values are to be expected and emitting a warning may be helpful.
- Using a newer version than downstream: the options may have changed in the update and old options could no longer be valid. Not sure what a good default behaviour would be.
- Using an older version than downstream: the user might be unaware of the update and we should warn him about the update and options, which a dependency might rely on.

In all cases we might consider adding an `OVERRIDE ON` parameter to disable the warning, as the user signals that he knows what he is doing

PR #244, opened 2021-04-15
Respect `FETCHCONTENT_BASE_DIR` if set by a user

trxcllnt:


cmake/CPM.cmake
```diff
@@ -525,7 +525,12 @@ function(CPMAddPackage)
     list(APPEND CPM_ARGS_UNPARSED_ARGUMENTS SOURCE_DIR ${download_directory})
     if(EXISTS ${download_directory})
       # avoid FetchContent modules to improve performance
-      set(${CPM_ARGS_NAME}_BINARY_DIR ${CMAKE_BINARY_DIR}/_deps/${lower_case_name}-build)
+      if(DEFINED FETCHCONTENT_BASE_DIR)
+        # respect FETCHCONTENT_BASE_DIR if set
+        set(${CPM_ARGS_NAME}_BINARY_DIR ${FETCHCONTENT_BASE_DIR}/${lower_case_name}-build)
+      else()
+        set(${CPM_ARGS_NAME}_BINARY_DIR ${CMAKE_BINARY_DIR}/_deps/${lower_case_name}-build)
+      endif()
```
TheLartians, line 533: How about merging the conditional here and below by defining and using a common variable `FETCH_CONTENT_ROOT` that is conditionally set to `${FETCHCONTENT_BASE_DIR}` or `${CMAKE_BINARY_DIR}/_deps` above? This would make it easier to use in other places and avoid future bugs.

TheLartians (commented):
Hey, thanks for the PR! It's a good idea to allow users to define custom FetchContent base directories. Imo it might be a good idea to merge the two branches to avoid repetition.

cmake/CPM.cmake
trxcllnt: Done!

trxcllnt:
> It's a good idea to allow users to define custom FetchContent base directories

This PR solves the case where if `CPM_SOURCE_CACHE` is unpopulated, dependencies are built under `FETCHCONTENT_BASE_DIR`, but if the cache is already populated they're built in `${CMAKE_BINARY_DIR}/_deps` .

TheLartians (approved):
Looks great, thanks for the changes!

TheLartians:
> This PR solves the case where if CPM_SOURCE_CACHE is unpopulated, dependencies are built under FETCHCONTENT_BASE_DIR, but if the cache is already populated they're built in ${CMAKE_BINARY_DIR}/_deps .

Shouldn't FetchContent itself respect `FETCHCONTENT_BASE_DIR` in that case?

trxcllnt:
It does, but in the case the cache is already populated CPM takes the branch with this comment:
> avoid FetchContent modules to improve performance

:upside_down_face:

TheLartians:
Ah yes, I think misunderstood your comment. 😅 All good then, thanks for clarifying!

TheLartians:
Fix is released in `v0.32.1` 🎉 
Thanks for the PR!

PR #252, opened 2021-05-14
Update boost-cmake example

TheLartians:
This should fix the broken travis build due to a hash mismatch.


TheLartians:
Welp, that fixed travis, but it seems that clang-format recently had a breaking change and there doesn't seem to be a good way to version-lock it through brew. 😒

iboB (approved):

Issue #255, opened 2021-05-28
[Feature request] Add bitbucket.org support


pgorgon-hem:
It would be nice to have same level of support of bitbucket.org repositories as of github or gitlab repos. I mean it needs to add variable BITBUCKET_REPOSITORY and short syntax "bb:org/repo@version". It looks easy to add and I'll make a PR for that.

pgorgon-hem:
It is added in #256

PR #256, opened 2021-05-28
Added support for bitbucket repositories

pgorgon-hem:
* added variable BITBUCKET_REPOSITORY
* added short syntax in form "bb:user/repo"
* added description of that to readme.md
Connected feature request: #255


TheLartians:
Hey, thanks for the PR! Could you add a test case for the shorthand syntax in [`test/unit/parse_add_package_single_arg.cmake`](https://github.com/cpm-cmake/CPM.cmake/blob/master/test/unit/parse_add_package_single_arg.cmake)?

Also it looks like CMake.org's SSL certificate has expired, which is why some of the test cases are failing. I'll check and relaunch the remaining tests later.

pgorgon-hem:
Hi! Of course I will add example and test. In case of cmake.org SSL certificate there is probably no choice and we have to wait for renew of cert.

iboB (dismissed):
Thanks

cmake/CPM.cmake
```diff
@@ -405,6 +409,10 @@ function(CPMAddPackage)
     set(CPM_ARGS_GIT_REPOSITORY "https://gitlab.com/${CPM_ARGS_GITLAB_REPOSITORY}.git")
   endif()
 
+  if(DEFINED CPM_ARGS_BITBUCKET_REPOSITORY)
+    set(CPM_ARGS_GIT_REPOSITORY "https://bitbucket.org/${CPM_ARGS_BITBUCKET_REPOSITORY}.git")
+  endif()
```
TheLartians, line 414: Perhaps we could group the three `if` clauses using `elseif` as well.

TheLartians (changes requested):
@pgorgon-hem thanks for adding the test! Could you still add a short example using a bitbucket repo for integration testing?

cmake/CPM.cmake
pgorgon-hem: I grouped it and pushed.

TheLartians (dismissed):
Thanks for the changes, looks great! Just to be sure, are you still planning on adding an example?

pgorgon-hem:
I added example repo on BB - our TestingFramework, but I'm not sure whether it compiles on server because of minimum cmake version required in our code.

TheLartians:
Yep, this definitely breaks CI, as it requires CMake 3.19 (we are still on 3.14). 

<details><summary>Error message</summary>
<p>

```
running example TestingFramework
1263================================
1264- cmake -Hexamples/TestingFramework -Bbuild/TestingFramework
1265error while running 'cmake -Hexamples/TestingFramework -Bbuild/TestingFramework':
1266   CMake Error at /home/travis/build/cpm-cmake/CPM.cmake/build/TestingFramework/_deps/testingframework-src/CMakeLists.txt:1 (cmake_minimum_required):
1267    CMake 3.19 or higher is required.  You are running version 3.14.0
1268
```

</p>
</details>

I've updated CMake to 3.20 in #264, and will let you know once you can merge!

pgorgon-hem:
Great, thanks! So I'm waiting for new cmake.

iboB:
The style still needs to be fixed, right?

Otherwise this works with #264 in

pgorgon-hem:
Now everything should be ok including styles (I tested it on our fork).

TheLartians (approved):
Looks great, thanks for adding the test and example!

TheLartians:
The changes are now published in `v0.32.2`! 🎉

PR #258, opened 2021-06-01
Fix typo in README - omit -> emit

claremacrae:


TheLartians:
Thanks for the check!

Issue #263, opened 2021-06-07
No automatic updates if GIT_TAG points to a branch 


DmitryYurov:
I am trying to use cpm for automating dependencies of my projects.
Unfortunately if I use cpm like
```
CPMAddPackage(
  NAME name_of_repo
  GIT_REPOSITORY ssh://git@server/project.git
  GIT_TAG master
  EXCLUDE_FROM_ALL YES
)
```
the fetched resource does not track the master branch - that is, on update of my master branch I still get older commit in the local repository. On the other hand, if I manually remove the fetched data, new download looks at the proper commit.

If that matters, I set `CPM_SOURCE_CACHE` environment variable.
Is it a bug or expected behaviour? I would expect that resources are downloaded again if the branch state in remote repository changes.

TheLartians:
This is expected behaviour, as CPM.cmake assumes that the `TAG` argument point to an unchangeable git tag or commit hash. We encourage this, as otherwise builds would become unstable, e.g. they would change every time the branch is modified. Also this allows us to efficiently cache sources and have a predictable offline and online behaviour. 

For parallel development of dependency / dependant projects, I would either manually update the tag to the most recent commit or use a [local package override](https://github.com/cpm-cmake/CPM.cmake#local-package-override) to pull the source from a local directory that is kept up to date with the master branch.

DmitryYurov:
In this case the functionality related to branch-named tag seems to be internally contradictory - on the one hand if I reference the tip of the branch, I would like the reference to stay on this tip - but it is fixed on the last commit at the moment of the first dependency download. On the other hand, this tag does not provide the stability, since the user could for example setup a clean build on the server and in such a way update the reference all the time. The behavior of fixed dependency in case of a version tag looks good, but with branch-named tag it is quite unexpected.

TheLartians:
You're right, the README is actually wrong in this case as it was written before we implemented caching and offline support. We actively discourage the use of branch names, as it would prevent us from supporting offline builds and reproducibility due to the reasons mentioned above. Perhaps we could add an alternative option `BRANCH_NAME` that would allow the old functionality of auto-updating branches.

N3m3si5:
[...]

PR #265, opened 2021-06-07
Remove branch name references from README

TheLartians:
As noted in #263, setting `GIT_TAG` to a branch name results in unexpected and unreproducible build behaviour, as it does not automatically update when the `HEAD` changes.


iboB:
Not sure whether just removing the refs is enough. Using a branch name still (kinda) works. Shouldn't there be a mention of that?

TheLartians:
Well I was thinking of adding an option `BRANCH_NAME` in the future to fall back to the old (non offline) behaviour for branches, and for now only encouraging the use of tags or commit hashs as branches lead to unexpected caching behaviour. But I see your point and will document it later!

pgorgon-hem:
> Well I was thinking of adding an option `BRANCH_NAME` in the future to fall back to the old (non offline) behaviour for branches, and for now only encouraging the use of tags or commit hashs as branches lead to unexpected caching behaviour. But I see your point and will update the PR later!

I really like the option BRANCH_NAME, it is probably same as TAG with NO_CACHE option, so looks easy to add.

iboB (approved):

PR #267, opened 2021-06-08
Expand relative download directory to prevent unnecessary download

studoot:
If `CMAKE_SOURCE_CACHE` (and thus `download_directory` in `CPMAddPackage`) is set to a relative path, the condition `if(EXISTS ${download_directory})` will fail even if the required directory exists. Expanding the variable `download_directory` to an absolute path before the test will rectify this, preventing unnecessary downloads of the sources.

### Motivation

I've looked at CPM (vs. vcpkg, conan and others) a couple of times before, and been put off by lengthy re-configure times. I've realised that this is because I've specified the cache directory as a relative path (e.g. with `set(CPM_SOURCE_CACHE ../.cpm_cache CACHE PATH "CPM download cache" FORCE)` in my main `CMakeLists.txt`). I could change *my* code, but I figure if it's done inside CPM, then others may also benefit.

I'll certainly be using CPM in the future now I've sussed why I was having issues previously.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -531,6 +531,8 @@ function(CPMAddPackage)
     list(SORT origin_parameters)
     string(SHA1 origin_hash "${origin_parameters}")
     set(download_directory ${CPM_SOURCE_CACHE}/${lower_case_name}/${origin_hash})
+    # Expand `download_directory` relative path. This is important because EXISTS doesn't work for relative paths.
+    get_filename_component(download_directory ${download_directory} ABSOLUTE)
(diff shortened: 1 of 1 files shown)


TheLartians (changes requested):
Hey @studoot, thanks for the insight and patch! Paths are tricky in CMake and it's great to support relative paths as well.

Could you run cmake-format on your code and push again? You can call the formatter through CMake as well:

```bash
cmake -S test/style -B build/style
```

studoot:
Those commands have been run & fixed the formatting.

TheLartians (approved):
Awesome, thanks!

TheLartians:
The changes are now published in `v0.32.2`! 🎉

Issue #266, opened 2021-06-08
Clion shows packages located in cache as hashes. There should be names


pgorgon-hem:
![image](https://user-images.githubusercontent.com/35261470/121159687-b0d84d00-c84b-11eb-91c2-870ebea75ca6.png)
I can propose quick fix in directories naming scheme, ie. add directory with package name inside hash-named directory. The problem is that it brakes current cache structure. I have 2 solutions: additional variable which enables this feature or detection of old cache and use it, copy to the new one or remove and populate again. What do you think?

pgorgon-hem:
Example of my solution: 
![image](https://user-images.githubusercontent.com/35261470/121160779-abc7cd80-c84c-11eb-8a3b-928fb4851112.png) 
and in code:
![image](https://user-images.githubusercontent.com/35261470/121160856-beda9d80-c84c-11eb-8c9c-50debf75eb30.png)

iboB:
I am generally NOT in favor of this for backwards compatibility reasons though it could work with this workaround.

I wonder whether we can request a CLion feature to somehow the name of such directories. It will also work with FetchContent-based projects which don't use CPM

pgorgon-hem:
It also sounds good, but I will make PR, you can look into code. I tried a few options and I think that movement of old cache directories to new location looks the best.

TheLartians:
Feature added in #268.

PR #268, opened 2021-06-08
Added directory inside cache to make it shown more friendly in CLion

pgorgon-hem:
Refering to #266 I changed directory structure of cache. To make it compatible with old cache I added logic moving old structure to new one.


TheLartians:
Hey @pgorgon-hem, thanks for the PR! I see how this would help organisation in IDEs such as CLion, but tbh I'm not a fan of moving directories around in the cache. Note that you can have multiple projects using different versions of CPM.camke, which would create re-downloads and introduce unexpected bugs into projects as well as the dependencies change location.

I don't use CLion, but maybe there is some other way to fix the directory naming there? 🤔

pgorgon-hem:
I absolutelly agree that moving directories may brake older builds, so maybe I can generate hash in another way and put new sources there, it should help. CLion uses [cmake-file-api](https://cmake.org/cmake/help/git-stage/manual/cmake-file-api.7.html) to get information about project. There are directories entries in reply and I suppose it gets information only from there. So CMake is a source of that information, not CLion itself.
image

TheLartians:
It's definitely a better approach, would result in duplicate dependency downloads, depending on the CPM version but that wouldn't be problematic imo. Perhaps in the initial phase we could make the new path opt-in by using a flag / env variable like `CPM_NAMED_CACHE_DIRECTORIES` or similar. 

@iboB any takes on this?

pgorgon-hem:
Agree. So I added CPM_USE_NAMED_CACHE_DIRECTORIES.

pgorgon-hem:
Hi, any progress?

TheLartians:
Hey, for some reason it seems that Travis is not reporting and I'm unable to restart the job... Could you try pushing an empty commit to trigger it?

pgorgon-hem:
I used oportunity and added description of that function to readme.

pgorgon-hem:
I see that travis still can't report. Additionally I don't know why style doesn't work, do you have any idea?

TheLartians:
@pgorgon-hem could you pull from master and try again? I've fixed the code creating styling problems and the checks should now all run on GitHub.

pgorgon-hem:
> @pgorgon-hem could you pull from master and try again? I've fixed the code creating styling problems and the checks should now all run on GitHub.

I've merged master and now it works!

TheLartians (approved):
Hey, sorry for the inactivity, I lost track of the PRs. Looks good, I still wonder if this is the best approach to fixing the naming, but I don't see any issue with including the feature!

TheLartians:
Released in `v0.34.0` 🎉

pgorgon-hem:
> Hey, sorry for the inactivity, I lost track of the PRs. Looks good, I still wonder if this is the best approach to fixing the naming, but I don't see any issue with including the feature!

Thx so much!

PR #284, opened 2021-08-27
Detecting when the cache is dirty 

andijcr:
#273 
I took the liberty to modify the fork of @apGribble to submit a pull request

this branch adds a check for uncommitted changes in the cache if the folder is a git repo, printing a warning.
in case the folder is not a git repo, a status message is printed and no real check is performed.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -354,6 +354,32 @@ function(cpm_parse_add_package_single_arg arg outArgs)
   )
 endfunction()
 
+# Check that the working directory for a git repo is clean
+function(cpm_check_git_working_dir_is_clean repoPath isClean)
+  # Not sure this check is necessary given git is probably key to the rest of the script
+  find_package(Git REQUIRED)
+
+  execute_process(
+    COMMAND ${GIT_EXECUTABLE} status --porcelain
+    RESULT_VARIABLE result
+    OUTPUT_VARIABLE status
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    WORKING_DIRECTORY ${repoPath}
+  )
+
+  if(result)
+    message(FATAL_ERROR "Calling git status on folder ${repoPath} failed")
+  endif()
+
+  if("${status}" STREQUAL "")
+    set(${isClean} TRUE PARENT_SCOPE)
+  else()
+    set(${isClean} FALSE PARENT_SCOPE)
+  endif()
+
+endfunction()
+
+
 # Download and add a package from source
 function(CPMAddPackage)
   list(LENGTH ARGN argnLength)
@@ -544,6 +570,13 @@ function(CPMAddPackage)
       set(${CPM_ARGS_NAME}_BINARY_DIR ${CPM_FETCHCONTENT_BASE_DIR}/${lower_case_name}-build)
       set(${CPM_ARGS_NAME}_ADDED YES)
       set(${CPM_ARGS_NAME}_SOURCE_DIR ${download_directory})
+
+      # warn if cache has been changed since checkout
+      cpm_check_git_working_dir_is_clean(${download_directory} IS_CLEAN)
+      if(NOT ${IS_CLEAN})
+        message(WARNING "Cache for ${CPM_ARGS_NAME} (${download_directory}) is dirty")
+      endif()
+
       cpm_add_subdirectory(
         "${CPM_ARGS_NAME}" "${DOWNLOAD_ONLY}"
         "${${CPM_ARGS_NAME}_SOURCE_DIR}/${CPM_ARGS_SOURCE_SUBDIR}" "${${CPM_ARGS_NAME}_BINARY_DIR}"
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -364,7 +364,7 @@ function(cpm_check_working_dir_is_clean repoPath isClean)
     RESULT_VARIABLE result
     OUTPUT_VARIABLE status
     OUTPUT_STRIP_TRAILING_WHITESPACE
-	ERROR_QUIET
+    ERROR_QUIET
     WORKING_DIRECTORY ${repoPath}
   )
 
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -355,41 +355,12 @@ function(cpm_parse_add_package_single_arg arg outArgs)
 endfunction()
 
 # Check that the working directory for a git repo is clean
-function(cpm_check_working_dir_is_clean repoPath gitTag isClean)
+function(cpm_check_git_working_dir_is_clean repoPath gitTag isClean)
 
   find_package(Git REQUIRED)
 
   if(NOT GIT_EXECUTABLE)
     # No git executable, assume directory is clean
-    set(${isClean}
-        TRUE
-        PARENT_SCOPE
-    )
-    return()
-  endif()
-
-  # get absolute path of .git dir, to find if we are in a top-level repo path
-  execute_process(
-    COMMAND ${GIT_EXECUTABLE} rev-parse --absolute-git-dir
-    RESULT_VARIABLE resultGitFolder
-    OUTPUT_VARIABLE gitFolderPath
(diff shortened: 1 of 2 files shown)


cmake/CPM.cmake
```diff
@@ -354,6 +354,33 @@ function(cpm_parse_add_package_single_arg arg outArgs)
...
+    RESULT_VARIABLE result
+    OUTPUT_VARIABLE status
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    ERROR_QUIET
+    WORKING_DIRECTORY ${repoPath}
+  )
+
+  # Not a git repo. for now, assume the directory is clean
+  if(result)
+    message(STATUS "not a git repo, can't check if clean:  ${repoPath}")
```
TheLartians, line 373: IMO this status message can be removed.

cmake/CPM.cmake
```diff
@@ -354,6 +354,33 @@ function(cpm_parse_add_package_single_arg arg outArgs)
   )
 endfunction()
 
+# Check that the working directory for a git repo is clean
+function(cpm_check_working_dir_is_clean repoPath isClean)
+  # Not sure this check is necessary given git is probably key to the rest of the script
+  find_package(Git REQUIRED)
```
TheLartians, line 360: Without the `find_package`, the variable `GIT_EXECUTABLE` would not be defined though - I think we can remove this comment.

cmake/CPM.cmake
```diff
@@ -354,6 +354,33 @@ function(cpm_parse_add_package_single_arg arg outArgs)
...
+
+  execute_process(
+    COMMAND ${GIT_EXECUTABLE} status --porcelain
+    RESULT_VARIABLE result
+    OUTPUT_VARIABLE status
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    ERROR_QUIET
+    WORKING_DIRECTORY ${repoPath}
+  )
```
TheLartians, line 369: I beliebe this can create false positives, e.g. if the dependency is non-git but stored inside an outer git repository. Perhaps we should only run the check if the repo itself is really a git directory, e.g. we could check if `${repoPath}/.git` exists.

TheLartians (commented):
Hey, thanks for the PR and helping to implement this feature!

cmake/CPM.cmake
andijcr: agreed, removed

cmake/CPM.cmake
andijcr: comment removed

cmake/CPM.cmake
andijcr: to work around i added git rev-parse --absolute-git-dir to get the location of .git, to compare it with ${repoPath}

andijcr:
I saw under test/unit that there is a cache.cmake test. should I add there a unit test for this feature?
I think a cmake script that creates a folder, inits a repo, then modifies a file.

andijcr:
added a new unit test dirty-cache-check.cmake

test/unit/dirty-cache-check.cmake
```diff
@@ -0,0 +1,58 @@
...
+git_do("${baseDir}" init -b main)
+message(STATUS "empty repo with file")
+cpm_check_working_dir_is_clean(${baseDir} emptygit_test)
+assert_falsy(emptygit_test)
+
+git_do("${baseDir}" add draft.txt)
+git_do("${baseDir}" commit -m "test change")
+message(STATUS "commit a change")
+cpm_check_working_dir_is_clean(${baseDir} onecommit_test)
+assert_truthy(onecommit_test)
```
TheLartians, line 38: This is an interesting case I didn't consider before - do we want to accept adding new commits as "clean"?

I think the expected behavior might be to also validate that the `GIT_TAG` is as defined `CPMAddPackage`.

cmake/CPM.cmake
```diff
@@ -354,6 +354,76 @@ function(cpm_parse_add_package_single_arg arg outArgs)
...
+    RESULT_VARIABLE resultGitStatus
+    OUTPUT_VARIABLE repoStatus
+    OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET
+    WORKING_DIRECTORY ${repoPath}
+  )
+  if(resultGitStatus)
+    # not supposed to happen
+    message(FATAL_ERROR "Calling git status on folder ${repoPath} failed")
+    return()
+  endif()
```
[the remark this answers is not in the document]
andijcr: ok, will do

test/unit/dirty-cache-check.cmake
andijcr: makes sense. @TheLartians where should I look to extract the ```GIT_TAG``` of the ```CPMAddPackage``` call?

TheLartians: Inside the CPMAddPackage call, the argument should be named `CPM_ARGS_GIT_TAG`, so I think you could just add an additional argument `GIT_TAG` to the `cpm_check_working_dir_is_clean` function and change the call to 

```cmake
       cpm_check_working_dir_is_clean(${download_directory} "${CPM_ARGS_GIT_TAG}" IS_CLEAN)
```

test/unit/dirty-cache-check.cmake
```diff
@@ -0,0 +1,58 @@
...
+include(${CPM_PATH}/testing.cmake)
+
+set(baseDir "${CMAKE_CURRENT_BINARY_DIR}/test_dirty_cache")
+set(childDir "${baseDir}/edgecase")
+
+find_package(Git REQUIRED)
+
+function(git_do dir)
+  execute_process(
+    COMMAND ${GIT_EXECUTABLE} ${ARGN}
```
TheLartians, line 11: Also note that the unit tests are failing due to git not being fully configured on CI. I think this could be easily fixed by including them in the git call:

```suggestion
    COMMAND ${GIT_EXECUTABLE}  -c user.name='User' -c user.email='user@email.org' ${ARGN}
```

cmake/CPM.cmake
```diff
@@ -544,6 +627,15 @@ function(CPMAddPackage)
       set(${CPM_ARGS_NAME}_BINARY_DIR ${CPM_FETCHCONTENT_BASE_DIR}/${lower_case_name}-build)
       set(${CPM_ARGS_NAME}_ADDED YES)
       set(${CPM_ARGS_NAME}_SOURCE_DIR ${download_directory})
+
+      if(DEFINED CPM_ARGS_GIT_TAG)
```
andijcr, line 631: This check implies that the download_directory passed to cpm_check_working_dir_is_clean is a git repo.
This in turn should remove the need to check for the edgecase of a download_directory in an unrelated parent git repo.

In this commit the check is still present though.

andijcr:
I've been using this fork for my project and so far so good. My daily setup is a Linux machine with git configured, a mix of .zip libs, third-party git libs, and in-house git libs.
in-house git libs are the main reason for this fork, and in this regard, I was able to catch a couple of instances of uncommitted changes in the cache that I overlooked.

open questions:
1. since I have to `if(DEFINED CPM_ARGS_GIT_TAG)` is it ok to assume that if CPM_ARGS_GIT_TAG is defined then I am for sure in a git repo library?
1.1 in this case there is no need for a check for the edge use case - a .zip lib inside an unrelated git repo. should I keep the check nonetheless, to be defensive?
1.2 should i rename back `cpm_check_working_dir_is_clean` to `cpm_check_git_working_dir_is_clean` to reflect the fact that it's in fact meat only for git?

cmake/CPM.cmake
TheLartians: That's a good point, we can probably safely remove the check. Lmk if you want to remove it in this PR.

TheLartians (approved):
Looks great, thanks for adding this feature!

andijcr:
@TheLartians  ok, I went ahead removed the check and renamed the function. 
I think work for now is done, I'll keep using my branch to test it. Let me know if I have to do anything.

TheLartians (approved):
Looks great, thanks!

PR #302, opened 2021-11-28
Allow passing lists in the options

### LINKED ISSUE #291: It's not possible to pass list as option

TheLartians:
As #301, but allows passing lists using quadruple escaped semicolons. The reason being that, as arguments are beeing parsed and passed to CMake functions, each substitutes one layer of the escapes (`a\\\\;b` (parsing) -> `a\\;b` (`cmake_parse_arguments`)  -> `a\;b` (extracting the option list) -> `a;b`).

Closes #301 
Closes #291


amirmasoudabdol:
Would it be possible to use `$<SEMICOLON>` instead of `\\\\;`?

TheLartians:
Hm I'm not sure, what advantages would the `$<SEMICOLON>` syntax bring over the escaped pattern?

PR #308, opened 2021-12-15
Pass DOWNLOAD_ONLY flag when CPM_pkg_SOURCE is set.

### LINKED ISSUE #307: DOWNLOAD_ONLY not works together with CPM_pkg_SOURCE
I'm using `DOWNLOAD_ONLY YES` flag to just download package, without calling `add_subdirectory()`.

This behaviour does not work if I manually set `CPM_pkg_SOURCE` variable to use existing package location during development. 

Internally, CPM recursivelly calls CPMAddPackage() inside `CPMAddPackage()` [here](https://github.com/cpm-cmake/CPM.cmake/blob/dad37fbe4e7a64c42918159402847939265bcda1/cmake/CPM.cmake#L545).

But argument `DOWNLOAD_ONLY` is not passed to this recursive call.

So, is this normal or bug?

As workaround I'm using unexisting `SOURCE_SUBDIR` to skip `add_subdirectory()` call.

hacker-cb:
Fix #307

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -548,6 +548,7 @@ function(CPMAddPackage)
       EXCLUDE_FROM_ALL "${CPM_ARGS_EXCLUDE_FROM_ALL}"
       OPTIONS "${CPM_ARGS_OPTIONS}"
       SOURCE_SUBDIR "${CPM_ARGS_SOURCE_SUBDIR}"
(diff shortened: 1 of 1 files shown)


TheLartians (approved):
Good catch, thanks!

TheLartians:
Fix is published in `v0.34.2`.

Issue #281, opened 2021-07-27
Weird issue with CPM_SOURCE_CACHE and FetchContent


mrexodia:
In the top-level project there is https://github.com/vtil-project/VTIL-Utils/blob/master/cmake/VTIL-Core.cmake:

```cmake
CPMAddPackage(
    NAME VTIL-Core
    GIT_REPOSITORY https://github.com/mrexodia/VTIL-Core
    GIT_TAG e6debb384cb218caff8cb94f50f00076b994a762
)
```

And after that it includes https://github.com/vtil-project/VTIL-Utils/blob/master/cmake/VTIL-NativeLifters.cmake:

```cmake
CPMAddPackage(
    NAME VTIL-NativeLifters
    GIT_REPOSITORY https://github.com/vtil-project/VTIL-NativeLifters
    GIT_TAG f8837ef4d2412a5a1c25f725f25685a23e1cf1a1
)
```

Inside the `VTIL-NativeLifters` repo there is a `FetchContent` call with the name `VTIL-Core`, The idea here is that you can pin a different version of `VTIL-Core` in your project that will then be used to compile the `VTIL-NativeLifters` library.

https://github.com/vtil-project/VTIL-NativeLifters/blob/master/Dependencies/CMakeLists.txt#L4:

```cmake
include(FetchContent)

message(STATUS "Fetching VTIL-Core (this might take a while)...")
FetchContent_Declare(
    VTIL-Core
    GIT_REPOSITORY https://github.com/vtil-project/VTIL-Core
    GIT_TAG        eeffb26b8b9ef7b4bd6ffbfc09f3133b70b36eda
    GIT_SHALLOW    false
)
FetchContent_MakeAvailable(VTIL-Core)
```

When using `CPM_SOURCE_CACHE` CMake errors out:

```
D:\CodeBlocks\VTIL-Utils\build>cmake ..
-- Building for: Visual Studio 16 2019
-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19042.
-- The C compiler identification is MSVC 19.29.30038.1
-- The CXX compiler identification is MSVC 19.29.30038.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30037/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
[...]

Issue #227, opened 2021-03-01
No SOURCE_DIR for packages added by multiple projects when using cache


erri120:
### The Problem

CPM does not set `SomePackage_SOURCE_DIR` (and possible other variables) when multiple sub-projects (using `add_subdirectory`) add the same package and a source cache is defined with `CPM_SOURCE_CACHE`.

In my case the package with problems was `StableCoder/cmake-scripts` (see [cmake/Tools.cmake](#cmake/Tools.cmake)) which contains useful scripts that I want to include and use. `StableCoder-cmake-scripts_SOURCE_DIR` was apprently set for when the first sub-project included the `Tools.cmake` file but every other sub-project was not this fortunate. The [logs](#Logs) show that CPM successfully adds the package and the first project was able to include its scripts using the `SOURCE_DIR` variable but the following ones failed.

### Project Structure

```
|   CMakeLists.txt
|
+---cmake
|       CPM.cmake
|       Tools.cmake
|
+---exe
|       CMakeLists.txt
|       main.cpp
|
+---lib
|       CMakeLists.txt
|       stuff.cpp
|       stuff.h
|
\---test
        CMakeLists.txt
        main.cpp
        something.cpp
```

#### CMakeLists.txt

```cmake
cmake_minimum_required(VERSION 3.17)

project(Project)

add_subdirectory(exe)
add_subdirectory(lib)
add_subdirectory(test)
```

#### exe/CMakeLists.txt

```cmake
cmake_minimum_required(VERSION 3.17)

project(splitcmake)

include(../cmake/CPM.cmake)
message("I'm exe/CMakeLists.txt")
include(../cmake/Tools.cmake)

add_executable(splitcmake main.cpp)

include_directories(${CMAKE_SOURCE_DIR}/lib)
target_link_libraries(splitcmake splitcmakeLib)

cxx_17()
```

`lib/CMakeLists.txt`:

```cmake
cmake_minimum_required(VERSION 3.17)

project(splitcmakeLib)

include(../cmake/CPM.cmake)
message("I'm lib/CMakeLists.txt")
include(../cmake/Tools.cmake)

add_library(splitcmakeLib stuff.cpp stuff.h)

cxx_17()
```

#### test/CMakeLists.txt

```cmake
cmake_minimum_required(VERSION 3.17)

project(splitcmakeTests)

include(../cmake/CPM.cmake)
message("I'm lib/CMakeLists.txt")
include(../cmake/Tools.cmake)

[...]

PR #314, opened 2021-12-21
Fix <package>_SOURCE_DIR and _BINARY_DIR handling when caching is active

DNKpp:
Today I struggled a bit with a project of mine, where I used ``CPMAddPackage`` on exactly the same dependency but on different hirarchies of my project. The first add was fine, but the second did neither set the ``<package>_SOURCE_DIR`` nor ``<package>_BINARY_DIR``.
There are some other issues around, that might be related to this topic (#227 #281 )

I did some code digging and this is what I think causes the issue:
The first ``CPMAddPackage`` runs until the end of the function, but because the package is already in the local cache, it will skip the fetch case (~Line 672). Because of this skip, ``FetchContent`` doesn't know about the package. When ``CPMAddPackage`` is called the next time on the same package, ``CPM`` already knows that the package has been added (around line 536) and trys to load the ``SOURCE_DIR`` and ``BINARY_DIR`` from the ``FetchContent``, but those have never been set, thus both variables will be empty.

My solution is to utilize the cmake cache mechanism, but I really don't know if thats the best way to do it. Feel free to correct me ;)

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -619,10 +619,8 @@ function(CPMAddPackage)
     get_filename_component(download_directory ${download_directory} ABSOLUTE)
     list(APPEND CPM_ARGS_UNPARSED_ARGUMENTS SOURCE_DIR ${download_directory})
     if(EXISTS ${download_directory})
-      # avoid FetchContent modules to improve performance
-      set(${CPM_ARGS_NAME}_BINARY_DIR ${CPM_FETCHCONTENT_BASE_DIR}/${lower_case_name}-build)
-      set(${CPM_ARGS_NAME}_ADDED YES)
-      set(${CPM_ARGS_NAME}_SOURCE_DIR ${download_directory})
+	  cpm_store_fetch_properties(${CPM_ARGS_NAME} "${download_directory}" "${CPM_FETCHCONTENT_BASE_DIR}/${lower_case_name}-build")
+	  cpm_get_fetch_properties("${CPM_ARGS_NAME}")
 
       if(DEFINED CPM_ARGS_GIT_TAG)
         # warn if cache has been changed since checkout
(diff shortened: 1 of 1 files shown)


OlivierLDff:
I like the fact that cache is used, because anyway dependency is global to every projects.

DNKpp:
If I think about it and I'm getting more and more used onto that topic, we could potentially completely get rid of the ``FetchContent`` functions, exept one ``FetchContent_Populate`` if we use cache variables. But I don't know if that's the direction CPM wants to go.

OlivierLDff:
Having compatibility with fetchcontent help to make gradually the change over dependencies

DNKpp:
> Having compatibility with fetchcontent help to make gradually the change over dependencies

If I understand the authors correctly, they try to avoid ``FetchContent`` functions as much as possible. If they wouldn't this bug has not been present (look at the previous code comment).

TheLartians:
Hey, thanks for the PR, this does look like a good fix to the problem! One thing: could you run `cmake-format` on the code to update the code style? Alternatively, you can run `cmake -S test/style -B build/style && cmake --build build/style --target fix-format` and push the changes.

TheLartians:
> If I understand the authors correctly, they try to avoid FetchContent functions as much as possible. If they wouldn't this bug has not been present (look at the previous code comment).

To elaborate, the reason for avoiding `FetchContent` is that the performance is extremely bad in common used CMake versions, potentially adding more than a second of overhead per dependency on each configuration. This would effectively prevent CPM.cmake from being used in projects with hundreds of dependencies. You can check [this blog post](https://ibob.bg/blog/2020/01/13/cmake-package-management/) to see more details and some benchmarks.

DNKpp:
> Hey, thanks for the PR, this does look like a good fix to the problem! One thing: could you run `cmake-format` on the code to update the code style? Alternatively, you can run `cmake -S test/style -B build/style && cmake --build build/style --target fix-format` and push the changes.

Hey, glad that you like it. I applied the format and pushed it. Hopefully I did it correctly :D

TheLartians (approved):
Looks good, thanks!

TheLartians:
Fix is published in `v0.34.3`!

PR #326, opened 2022-01-11
Unified GitHub workflow for tests on major operating systems

iboB:
I think there is no point in having three separate workflows which are almost exactly the same. 

`ubuntu-latest` is Ubuntu 20.04 with CMake 3.22 so there is no need to download 3.14 explicitly (it used to be Ubuntu 18.04 with CMake 3.10)


iboB:
Note that there is no "build" action in this PR and this check will never complete. 

If this is merged, some repo configuration will be needed to remove "build" and set "tests" as a required job

TheLartians (approved):
Agreed. I think this came from a time where different build steps were necessary for each system, but it seems to have converged since then. Thanks!

PR #327, opened 2022-01-11
Remove badges

TheLartians:
As we now use the unified CI workflow (#326), it seems we unfortunately [can't have separate badges for each OS](https://github.community/t/separate-workflow-badges-when-using-matrix-testing-possible/16708). As the display of the badges was to reference the cross-platform nature of the project, imo it can be replaced by adding a reference in the introduction instead.


iboB (commented):
Perhaps we can use this instead https://github-actions.40ants.com/?url=https%3A%2F%2Fgithub.com%2Fcpm-cmake%2FCPM.cmake

TheLartians:
Yeah I had considered this as well, however I found the aesthetic rather distracting.

iboB (approved):

PR #329, opened 2022-01-12
[Example] Disable JSON tests being included in ctest

Vinpasso:
If `CPMAddPackage` includes json without setting JSON_BuildTests off first then the target's ctest will include lots of json's tests. This is probably undesired for folks that just want to use json to build their code.

### DIFF
--- a/examples/json/CMakeLists.txt
+++ b/examples/json/CMakeLists.txt
@@ -5,6 +5,8 @@ project(CPMJSONExample)
 # ---- Dependencies ----
 
 include(../../cmake/CPM.cmake)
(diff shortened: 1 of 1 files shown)


TheLartians:
Hey thanks for the PR! This [looks to be fixed](https://github.com/nlohmann/json/blob/4f8fba14066156b73f1189a2b8bd568bde5284c5/CMakeLists.txt#L33) in more recent versions of the library. If this works for you, let's update the example's version instead, e.g. to `3.10.5`.

Vinpasso:
Makes sense. I adjusted the version and undid the option set

TheLartians (approved):
Thanks for the update!

PR #335, opened 2022-01-26
Allows set/set(CACHE) to behave the same as set(X)/option(x)

robertmaynard:
By enabling CMP0126 we tell CMake that when constructing a CACHE variable
with the same name as a local variable to not unset the local variable.

By enabling CMPO126 it makes sure that CPM behavior around `OPTION` is
consistent if the option is created with `set(CACHE)` or with `option`.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -841,6 +841,12 @@ function(
       cmake_policy(SET CMP0077 NEW)
       set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
 
+      # the policy allows us to change set(CACHE) without caching
+      if(POLICY CMP0126)
+        cmake_policy(SET CMP0126 NEW)
+        set(CMAKE_POLICY_DEFAULT_CMP0126 NEW)
+      endif()
+
       foreach(OPTION ${OPTIONS})
         cpm_parse_option("${OPTION}")
         set(${OPTION_KEY} "${OPTION_VALUE}")


cmake/CPM.cmake
```diff
@@ -5,7 +5,7 @@
 # MIT License
 # -----------
 #[[
-  Copyright (c) 2021 Lars Melchior and additional contributors
+  Copyright (c) 2021 Lars Melchior, NVIDIA CORPORATION, and additional contributors
```
robertmaynard, line 8: I didn't see a place that listed contributors, so I added NVIDIA here.
If you don't want I can drop

iboB:
I'm generally in favor of the policy being set, but not of adding NVIDIA to the copyright in CPM.cmake. If we were to add every single contributor, the copyright notice could end up being longer than the actual code.

Still, having an external file which lists all contributors is probably not a bad idea

robertmaynard:
> I'm generally in favor of the policy being set, but not of adding NVIDIA to the copyright in CPM.cmake. If we were to add every single contributor, the copyright notice could end up being longer than the actual code.
> 
> Still, having an external file which lists all contributors is probably not a bad idea

I have removed that modification while you work out a better way to list contributors. Thank you

iboB (approved):

TheLartians:
Hey, thanks for the PR, setting the policy definitely looks like a good idea! As for the contributors, I'd be interested how other projects handle this. In general, I wouldn't be opposed to adding a contributors file, where we can give credit to the people / organisations that have helped with the project.

Issue #280, opened 2021-07-23
CPM_USE_LOCAL_PACKAGES for only one package


OlivierLDff:
Hi thank you for the great work
Would you consider a feature to be able to set `CPM_USE_LOCAL_PACKAGES` for only one package.
Like FetchContent:
> In addition to the above cache variables, the following cache variables are also defined for each content name (<ucName> is the uppercased value of <name>):
>
> FETCHCONTENT_SOURCE_DIR_<ucName>
> If this is set, no download or update steps are performed for the specified content and the <lcName>_SOURCE_DIR variable returned to the caller is pointed at this location. This gives developers a way to have a separate checkout of the content that they can modify freely without interference from the build. The build simply uses that existing source, but it still defines <lcName>_BINARY_DIR to point inside its own build area. Developers are strongly encouraged to use this mechanism rather than editing the sources populated in the default location, as changes to sources in the default location can be lost when content population details are changed by the project.
>
> FETCHCONTENT_UPDATES_DISCONNECTED_<ucName>
> This is the per-content equivalent of FETCHCONTENT_UPDATES_DISCONNECTED. If the global option or this option is ON, then updates will be disabled for the named content. Disabling updates for individual content can be useful for content whose details rarely change, while still leaving other frequently changing content with updates enabled.

So introducing `CPM_USE_LOCAL_PACKAGES_<ucName>` (or better but more error prone to copy and paste `CPM_USE_LOCAL_PACKAGE_<ucName>`).

I guess modification must be done around: 
https://github.com/cpm-cmake/CPM.cmake/blob/91585e3864cc0e2a140c70a57a05b76c12934419/cmake/CPM.cmake#L495-L509

```cmake
string(TOUPPER ${CPM_ARGS_NAME} UC_CPM_ARGS_NAME)
set(CPM_USE_LOCAL_PACKAGE_UC CPM_USE_LOCAL_PACKAGE_${UC_CPM_ARGS_NAME})
if(CPM_USE_LOCAL_PACKAGES OR CPM_LOCAL_PACKAGES_ONLY OR CPM_USE_LOCAL_PACKAGE_UC)
```

> I haven't tested this code yet

Would you consider such PR?

bilke:
Also the reverse, forcing a single library to be downloaded and handled by CPM, would be useful, e.g. `CPM_DOWNLOAD_<package_name>`.

TheLartians:
Hey thanks for the suggestions!

[...]

PR #336, opened 2022-01-26
Add per package CPM_DOWNLOAD controls

robertmaynard:
Introduces support for `CPM_<PACKAGE>_DOWNLOAD` variable ( and env ) which allows finer grained control.

A common workflow pattern we have run into is wanting to set `CPM_DOWNLOAD_ALL` to on for a middle package in a CPM dependency chain.

```
Root
  -> CPM project A
       -> CPM project B 
       -> CPM project C
```    
This would allow us to state that project A needs to be downloaded but project B or C can be found on disk.


README.md
```diff
@@ -183,6 +183,7 @@ You can use `CPM_SOURCE_CACHE` on GitHub Actions workflows [cache](https://githu
 If set, CPM will forward all calls to `CPMFindPackage` as `CPMAddPackage`.
 This is useful to create reproducible builds or to determine if the source parameters have all been set correctly.
 This can also be set as an environmental variable.
+This can be controlled on a per project basis with the `CPM_<dependency name>_DOWNLOAD` variable.
```
robertmaynard, line 186: I didn't see a place that listed contributors, so I added NVIDIA here.
If you don't want I can drop

iboB, line 186: "per package" as opposed to "per project" is probably better

cmake/CPM.cmake
```diff
@@ -5,7 +5,7 @@
 # MIT License
 # -----------
 #[[
-  Copyright (c) 2021 Lars Melchior and additional contributors
+  Copyright (c) 2021 Lars Melchior, NVIDIA CORPORATION, and additional contributors
```
iboB, line 8: As for your previous PR, I don't think adding individual contributors here is a good idea.

I will work on adding a separate list of contributors in the near future

cmake/CPM.cmake
```diff
@@ -251,7 +251,14 @@ function(CPMFindPackage)
...
   endif()
 
-  if(CPM_DOWNLOAD_ALL)
+  set(download_only ${CPM_DOWNLOAD_ALL})
+  if(DEFINED ENV{CPM_${CPM_ARGS_NAME}_DOWNLOAD})
+    set(download_only $ENV{CPM_${CPM_ARGS_NAME}_DOWNLOAD})
+  endif()
+  if(DEFINED CPM_${CPM_ARGS_NAME}_DOWNLOAD)
+    set(download_only ${CPM_${CPM_ARGS_NAME}_DOWNLOAD})
+  endif()
```
iboB, line 260: kinda nitpicky but since local variable trumps env which trumps global perhaps just do:

```cmake
if(DEFINED local)
  # ...
elseif(DEFINED env)
  # ...
else()
  set(download_only ${CPM_DOWNLOAD_ALL})
endif()
```
This makes the priority clearer, no?

cmake/CPM.cmake
```diff
@@ -251,7 +251,14 @@ function(CPMFindPackage)
     endif()
   endif()
 
-  if(CPM_DOWNLOAD_ALL)
+  set(download_only ${CPM_DOWNLOAD_ALL})
```
iboB, line 254: the convention for local vars is `camelCase`. Also tne name "download only" can be confused with the case where a package is downloaded but not configured. Why not use something like `downloadPackage`?

README.md
iboB, line 186: also I'm not sure about the name `CPM_<dep>_DOWNLOAD`. 

`CPM_Boost_DOWNLOAD` doesn't sound like proper English.

How about `CPM_DOWNLOAD_<dep>` 

"Hey, CPM, download this package!"

iboB (changes requested):
Thanks for the PR

OlivierLDff:
Interesting this like the variant I thought here https://github.com/cpm-cmake/CPM.cmake/issues/280 (but still didn't take time to implement)

README.md
robertmaynard: Removed since the project is looking at a better way to list contributors.

robertmaynard: That makes alot of sense as it also better matches `CPM_DOWNLOAD_ALL` (Hopefully we never encounter a package named ALL ).

cmake/CPM.cmake
robertmaynard: done

cmake/CPM.cmake
robertmaynard: Yes that is better.

iboB (approved):
I can't think of how this can break anything. Looks good to me. 

@TheLartians what do you think?

Also, this makes is a good use case for an integration test, I'll try do add one for this over the weekend

TheLartians:
Agreed, this looks good! Definitely worth it to add an integration test, so we don't accidentally break it in the future.

TheLartians:
@iboB are you still planning to write some integration tests for this feature?

iboB:
Yes. At least a very bare-bones one. I'm a bit short on time, though. I think if we're satisfied with this, we can merge and I can add tests later

TheLartians (approved):
Alright, thanks @robertmaynard for the feature!

PR #337, opened 2022-01-27
Use CMake 3.14+ documented way to pass the source dir `-S`.

robertmaynard:
The `-H<src_dir>` was never officially documented by CMake


iboB (approved):
Thanks

TheLartians (approved):
Makes sense, I was using `-H` in the past before I knew `-S` existed, but it's definitely time to switch to the official solution.

PR #338, opened 2022-01-27
Fix beginner’s trap / fix broken example for nlohmann/json

Tobi823:
The example for nlohmann_json is wrong. It's missing the download source.

Without this fix, my builds fails with this message: `No download info given for 'nlohmann_json-populate' and its source
  directory:`

Even the maintainer of nlohmann_json says (https://github.com/nlohmann/json#package-managers) that it should be:
```
CPMAddPackage(
    NAME nlohmann_json
    VERSION 3.9.1)
```


TheLartians (approved):
Good catch, thanks!

PR #340, opened 2022-01-29
List of contributors 

### LINKED ISSUE #339: List of contributors
This was raised as an issue in #335:  There is no list of individual contributors associated with the project. Such lists exist in many open source projects

For example here's what the best documented (IMO) open source project does: https://doc.magnum.graphics/magnum/credits-contributors.html

Doing something similar in terms of content is a lot of work. We would have to go through every single PR in order to match a contributor with what they've done.

So, instead I suggest we add `CONTRIBUTORS.md` where we list people from [this page](https://github.com/cpm-cmake/CPM.cmake/graphs/contributors) with links to their GitHub profiles and take it from there.

What to do with companies? If someone does contributions on company time, they would likely want to add the company as a contributor, as it seems to be the case with @robertmaynard 

@robertmaynard, how about we have your entry in the future `CONTRIBUTORS.md` look like this:

*  NVIDIA CORPORATION via [Robert Maynard](https://github.com/robertmaynard)

iboB:
Closes #339 

Also updated copyright years.

Order of list is based on a snapshot of https://github.com/cpm-cmake/CPM.cmake/graphs/contributors 

The order in this page changes upon reloads. The contributors are sorted by number of commits, but ones which have the same number of commits appear in different orders when reloading the page


TheLartians:
Thanks for creating the List! Do you think it would make sense to also link to the contributors file from the MIT licence?

data-man:
https://github.com/all-contributors/all-contributors? :)

iboB:
Let's hold off merging this. Some automation is needed here. Manually supporting this list is not a good idea. Perhaps the bot suggested by @data-man. I'll check it out tomorrow

iboB:
OK. So I no existing automated solution seems to do the job for us. The critical feature with our case is that we want to credit companies (as requested for NVIDIA) and not just individuals. 

So, I suggest we merge this as it is, and I'll try to find some time next week to automate this using the GitHub API. It doesn't look like something too complicated.

iboB:
> Thanks for creating the List! Do you think it would make sense to also link to the contributors file from the MIT license?

I don't think that's very important. Also GitHub license detection is pretty fragile (though in some might say that it's [not fragile enough](https://github.com/licensee/licensee/issues/236)) and adding such text will likely cause it to stop detecting the license as MIT which will make the info-box of the project have no license which does not look good.

TheLartians (approved):
Yeah, makes sense. I think this is a great start in any way.

Issue #328, opened 2022-01-12
CPM_INDENT is not inserted everywhere and no CPM_DESINDENT


flagarde:
Hello,

Thx for you nice tool it's wonderful.
I noticed that `CPM_INDENT` is not used everywhere in the code? This is expected ?

I would like to add an `CPM_DESINDENT` or other name to be able to use CPM_INDENT to use colorization. For now I can do it but I can't desactivate the color.

It wouyld be then possible to do : 
```cmake
set(CPM_INDENT MyColor)
set(CPM_DESINDENT White)
```
 
Would you accept a PR to add this variable `CPM_DESINDENT` or a better name ?

TheLartians:
Hey thanks for the feedback, happy you like it!

> I noticed that CPM_INDENT is not used everywhere in the code? This is expected ?

Hm I think there are some places where the indent wouldn't make sense as they are not scoped, e.g. fatal errors. For other purposes this may be an oversight.

> I would like to add an CPM_DESINDENT or other name to be able to use CPM_INDENT to use colorization. For now I can do it but I can't desactivate the color.

Hm not sure if that would be too far off from its original meaning. Maybe it would be better to introduce two new variables like `CPM_LEFT_LOG`, `CPM_RIGHT_LOG`? Probably it would also make sense to add a function for logging, for more consistency and easier customisation.

flagarde:
Yes I agree the name is not very nice. And the ones proposed are much better.. 

Concerning the indent on fatal errors it's true indentation are not useful but I used CPM_INDENT as [CPM]. 

Addind the values you propose looks great. For a logger I wrote something like this : https://github.com/flagarde/CMakeCM/blob/main/modules/Missives.cmake but it's maybe too much code just for logging some message in cpm

flagarde:
Would you allow a PR with CPM_LEFT_LOG CPM_RIGHT_LOG ?

TheLartians:
Sure, though if its a lot of extra code, I think I'd prefer a small helper function for logging. But I'd be happy to take a look at the PR!

flagarde:
Well it depends a lot of what you expect from a logger.

* Colorization : Possible but need to hack a bit as CMake doesn't provide any hint to know if the terminal or IDE accept ANSI color escape sequences. I raised an issue https://gitlab.kitware.com/cmake/cmake/-/issues/22954 for this.
* Add level : be able to register new levels based on the default ones.

[...]

PR #342, opened 2022-02-10
Add  CPM_LEFT_LOG and CPM_RIGHT_LOG #328

flagarde:
Add CPM_LEFT_LOG and CPM_RIGHT_LOG to allow user to customize them and use color for example.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -35,10 +35,10 @@ if(CPM_DIRECTORY)
     if(CPM_VERSION VERSION_LESS CURRENT_CPM_VERSION)
       message(
         AUTHOR_WARNING
-          "${CPM_INDENT} \
+          "${CPM_LEFT_LOG}${CPM_INDENT} \
 A dependency is using a more recent CPM version (${CURRENT_CPM_VERSION}) than the current project (${CPM_VERSION}). \
 It is recommended to upgrade CPM to the most recent version. \
-See https://github.com/cpm-cmake/CPM.cmake for more information."
+See https://github.com/cpm-cmake/CPM.cmake for more information.${CPM_RIGHT_LOG}"
       )
     endif()
     if(${CMAKE_VERSION} VERSION_LESS "3.17.0")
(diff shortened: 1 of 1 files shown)


TheLartians (commented):
Hey thanks for the PR! If I understand correctly from the issue, the main use-case is to add user-defined formatting to CPM.cmake logs?

How about adding two functions `cpm_log(message)` and `cpm_warning(message)` for logging. First of all that way there would be less code repetition and the formatting would be consistent. Also I think that those can be easily overridden by users for special formatting.

cmake/CPM.cmake
```diff
@@ -210,11 +210,25 @@ if(NOT CPM_INDENT)
   )
 endif()
 
+if(NOT CPM_LEFT_LOG)
```
CraigHutchinson, line 213: +1 for the suggestion to use `cpm_message()` and allow specialization on that in a single location. 

However, on keeping this approach the choise of variables naming is odd imho. I would prefer `CPM_LOG_PREFIX` and `CPM_LOG_POSTFIX` to be more self explanatory on the intent.

CraigHutchinson:
An alternative approach may resolve around a wider refactor of Log messages to use the 'CHECK_START' type messages and therefore the pre-existing [CMAKE_MESSAGE_INDENT]mechanism(https://cmake.org/cmake/help/latest/variable/CMAKE_MESSAGE_INDENT.html#variable:CMAKE_MESSAGE_INDENT)  can be user-customized

TheLartians:
Agree completely with most of your points, except for how easy updating the version of CMake is. From my experience there are quite a few companies where updating packages from a linux distribution is quite complicated, so they are stuck with very old software.

As you say, without analytics / surveys etc. it's very hard to estimate how many users this will impact. Perhaps instead of a "Next" version, we could maintain a legacy branch for old CMake versions that still get's bug and security fixes but no features that require newer CMake. The download count of that branch can also give us information on the demand for old version support.

flagarde:
> An alternative approach may resolve around a wider refactor of Log messages to use the 'CHECK_START' type messages and therefore the pre-existing [CMAKE_MESSAGE_INDENT](https://cmake.org/cmake/help/latest/variable/CMAKE_MESSAGE_INDENT.html#variable:CMAKE_MESSAGE_INDENT) that can be user-customized
> 
> EDIT: There is also [CMAKE_MESSAGE_CONTEXT](https://cmake.org/cmake/help/latest/variable/CMAKE_MESSAGE_CONTEXT.html#variable:CMAKE_MESSAGE_CONTEXT) which can be enabled by users via `--log-context` or [CMAKE_MESSAGE_CONTEXT_SHOW](https://cmake.org/cmake/help/latest/variable/CMAKE_MESSAGE_CONTEXT_SHOW.html#variable:CMAKE_MESSAGE_CONTEXT_SHOW) which may satisfy the use-case of the OP?

This is a bit orthogonal, I wanted to add color to CPM message so I needed to have the possibility to add a variable before AND after message to switch back to the default color. I have created some function for this but I needed to change the function inside CPM 

https://github.com/flagarde/CMakeCM/blob/main/modules/Missives.cmake
Screenshot_20221129_200009

PR #346, opened 2022-04-07
Add liblava to "Built with CPM.cmake"

TheLavaBlock:
We recently fully switched to CPM.cmake. Thanks!


TheLartians (approved):
Thanks, happy to have you aboard!

PR #363, opened 2022-06-20
Klogg switched to CPM for dependency management

variar:
Klogg used to vendor all external dependencies, and now we switched to CPM, and this makes using local dependencies and packaging much easier. Thanks for all you effort!


TheLartians (approved):
Thanks, looks like a very useful project! Happy to help out.

PR #364, opened 2022-06-23
Instead of spdlog's bundled fmt re-use it from CPM.cmake

jhasse:
I think this is a good example. If someone wants to stick to the bundled one it's easier to remove the OPTION than to add it otherwise.

### DIFF
--- a/examples/spdlog/CMakeLists.txt
+++ b/examples/spdlog/CMakeLists.txt
@@ -8,15 +8,11 @@ include(../../cmake/CPM.cmake)
 
 CPMAddPackage("gh:gabime/spdlog@1.8.2")
 
-# spdlog uses fmt and bundles that dependency. If you want to use fmt in your 
-# project as well, you can let spdlog re-use fmt from CPM.cmake like this:
+# spdlog uses fmt and bundles that dependency. If you want to use fmt in your project as well, you
+# can let spdlog re-use fmt from CPM.cmake like this:
 #
-# CPMAddPackage("gh:fmtlib/fmt#7.1.3")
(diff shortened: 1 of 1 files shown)


TheLartians (commented):
Hey, thanks for the PR! I see how this can be helpful for projects using both libraries.
However, for the sake of a minimal example I think I prefer keeping the one-liner here. How about additionally adding the explicit `fmt` example below in a comment with a short explanation instead?

jhasse:
like this?

examples/spdlog/CMakeLists.txt
```diff
@@ -8,6 +8,16 @@ include(../../cmake/CPM.cmake)
 
 CPMAddPackage("gh:gabime/spdlog@1.8.2")
 
+# spdlog uses fmt and bundles that dependency. If you want to use fmt anyway,
+# you can let spdlog re-use fmt from CPM.cmake like this:
```
TheLartians, line 12: I feel this rephrase makes it clearer to the reader what's happening.

```suggestion
# spdlog uses fmt and bundles that dependency. If you want to use fmt in your 
# project as well, you can let spdlog re-use fmt from CPM.cmake like this:
```

TheLartians (commented):
Exactly! Though maybe we can make the comment a little more descriptive as suggested.

TheLartians (approved):
Thanks for the changes!

TheLartians (dismissed):

TheLartians (approved):
CMake-format screwed up the code example, I had to disable formatting for that part.

PR #370, opened 2022-07-11
CPMAddPackage fails if the SOURCE_DIR directory is deleted.

thomas-mp:
When using CPMAddPackage with the SOURCE_DIR attribute.

If the SOURCE_DIR directory is deleted after a valid build CPM.cmake is not able to recover and re-fetch the content.

But it will leave the user in a bad state where the WORKING_DIRECTORY is invalid and it will abort after it has called the git stash in the ROOT of the source directory with the:
`git stash save --quiet;--include-untracked `

I order for the user to fix this:
`git stash pop`

This patch will check if the SOURCE_DIR really exists and if not it will delete the -subbuild directory in order to re-fetch the package.

Tested with: Pop!_OS 22.04 LTS (cmake version 3.22.2)


cmake/CPM.cmake
```diff
@@ -650,6 +650,18 @@ function(CPMAddPackage)
     list(APPEND CPM_ARGS_UNPARSED_ARGUMENTS DOWNLOAD_COMMAND ${CPM_ARGS_DOWNLOAD_COMMAND})
   elseif(DEFINED CPM_ARGS_SOURCE_DIR)
     list(APPEND CPM_ARGS_UNPARSED_ARGUMENTS SOURCE_DIR ${CPM_ARGS_SOURCE_DIR})
+    if(NOT IS_ABSOLUTE ${CPM_ARGS_SOURCE_DIR})
+      # Expand `CPM_ARGS_SOURCE_DIR` relative path. This is important because EXISTS doesn't work for
+      # relative paths.
+      get_filename_component(source_directory ${CPM_ARGS_SOURCE_DIR} REALPATH BASE_DIR ${CMAKE_BINARY_DIR})
```
TheLartians, line 656: I think it should be `CMAKE_CURRENT_BINARY_DIR`, as according to the [CMake docs](https://cmake.org/cmake/help/latest/module/ExternalProject.html):

> Relative paths are interpreted with respect to [CMAKE_CURRENT_BINARY_DIR](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_BINARY_DIR.html#variable:CMAKE_CURRENT_BINARY_DIR) at the point where ExternalProject_Add() is called.

TheLartians:
Thanks for the PR! Seems like a good idea to handle this edge-case. Two things:

- Could you add an [integration test](https://github.com/cpm-cmake/CPM.cmake/tree/master/test/integration) to test and preserve the functionality? [This one](https://github.com/cpm-cmake/CPM.cmake/blob/master/test/integration/test_fetchcontent_compatibility.rb) might be a good starting reference.
- Can you use cmake-format to auto-style the code? E.g. you can simply run:

```bash
```

TheLartians (changes requested):
Hey, thanks for the PR! Seems like a good idea to handle this edge-case. Two things:

- Could you add an [integration test](https://github.com/cpm-cmake/CPM.cmake/tree/master/test/integration) to test and preserve the functionality? [This one](https://github.com/cpm-cmake/CPM.cmake/blob/master/test/integration/test_fetchcontent_compatibility.rb) might be a good starting reference.
- Can you use cmake-format to auto-style the code? You could run it using the CMake target:

```bash
```

cmake/CPM.cmake
thomas-mp: Thanks, this is fixed now.

thomas-mp:
> Hey, thanks for the PR! Seems like a good idea to handle this edge-case. Two things:
> 
> * Could you add an [integration test](https://github.com/cpm-cmake/CPM.cmake/tree/master/test/integration) to test and preserve the functionality? [This one](https://github.com/cpm-cmake/CPM.cmake/blob/master/test/integration/test_fetchcontent_compatibility.rb) might be a good starting reference.
> * Can you use cmake-format to auto-style the code? You could run it using the CMake target:
> 
> ```shell
> cmake -Stest/style -Bbuild/style
> cmake --build build/style --target fix-format
> ```
Thanks for the starting reference. I have created the integration test and fixed the style.

test/integration/test_fetchcontent_source_dir.rb
```diff
@@ -0,0 +1,43 @@
...
+
+# Tests FetchContent overriding with CPM
+
+class FetchContentSourceDir < IntegrationTest
+  def setup
+    @cache_dir = File.join(cur_test_dir, 'cpmcache')
+    ENV['CPM_SOURCE_CACHE'] = @cache_dir
+  end
+
+  def test_add_dependency_cpm_and_fetchcontent
```
TheLartians, line 11: The function's name should also be changed

```suggestion
  def test_fetchcontent_source_dir
```

Though perhaps a more descriptive name for this and the test file would be better? Maybe something like "test_remove_source_dir" / "RemoveSourceDir"?

test/integration/test_fetchcontent_source_dir.rb
```diff
@@ -0,0 +1,43 @@
+require_relative './lib'
+
+# Tests FetchContent overriding with CPM
```
TheLartians, line 3: Comment should be updated or removed.

test/integration/test_fetchcontent_source_dir.rb
```diff
@@ -0,0 +1,43 @@
...
+
+class FetchContentSourceDir < IntegrationTest
+  def setup
+    @cache_dir = File.join(cur_test_dir, 'cpmcache')
+    ENV['CPM_SOURCE_CACHE'] = @cache_dir
+  end
```
TheLartians, line 9: Can probably be removed as not used in the test.

```suggestion
```

TheLartians (changes requested):
Thanks for adding the test! Some small final requests.

thomas-mp:
Nice thanks. We are getting there :)

TheLartians (approved):
Looks good, thanks for the changes!

PR #372, opened 2022-07-14
Timestamps for URL downloads match the download time

robertmaynard:
By enabling CMake policy 135 we ensure that extracted files timestamp match that of the download time, instead of when the archive is created. This makes sure that if the URL changes to an older version we still rebuild everything as the timestamp
stays newer.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -115,7 +115,6 @@ set(CPM_SET_RECOMMENDED_CMAKE_POLICIES
 
 if(CPM_SET_RECOMMENDED_CMAKE_POLICIES)
   # the policy allows us to change options without caching
-  cmake_policy(PUSH)
   cmake_policy(SET CMP0077 NEW)
   set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
 
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -595,13 +595,6 @@ function(CPMAddPackage)
     endif()
 
     list(APPEND CPM_ARGS_UNPARSED_ARGUMENTS URL "${CPM_ARGS_URL}")
-
-    # Prefer to use the download time for timestamp, instead of the timestamp in the archive unless
-    # explicitly set by user. This allows for proper rebuilds when a projects url changes
-    if(POLICY CMP0135)
-      cmake_policy(SET CMP0135 NEW)
-      set(CMAKE_POLICY_DEFAULT_CMP0135 NEW)
-    endif()
   endif()
 
   # Check for required arguments
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -108,6 +108,10 @@ set(CPM_DRY_RUN
     OFF
     CACHE INTERNAL "Don't download or configure dependencies (for testing)"
 )
+set(CPM_SET_RECOMMENDED_CMAKE_POLICIES
+    OFF
+    CACHE INTERNAL "Have CPM enable all recommended CMake policies"
+)
 
 if(DEFINED ENV{CPM_SOURCE_CACHE})
   set(CPM_SOURCE_CACHE_DEFAULT $ENV{CPM_SOURCE_CACHE})
@@ -239,8 +243,30 @@ function(cpm_create_module_file Name)
   endif()
 endfunction()
 
+macro(cpm_set_cmake_policies)
+  if(CPM_SET_RECOMMENDED_CMAKE_POLICIES)
+    # the policy allows us to change options without caching
+    cmake_policy(SET CMP0077 NEW)
+    set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
+
+    # the policy allows us to change set(CACHE) without caching
+    if(POLICY CMP0126)
+      cmake_policy(SET CMP0126 NEW)
+      set(CMAKE_POLICY_DEFAULT_CMP0126 NEW)
+    endif()
+
+    # The policy uses the download time for timestamp, instead of the timestamp in the archive.
+    # This allows for proper rebuilds when a projects url changes
+    if(POLICY CMP0135)
+      cmake_policy(SET CMP0135 NEW)
+      set(CMAKE_POLICY_DEFAULT_CMP0135 NEW)
+    endif()
+  endif()
+endmacro()
+
 # Find a package locally or fallback to CPMAddPackage
 function(CPMFindPackage)
+  cpm_set_cmake_policies()
   set(oneValueArgs NAME VERSION GIT_TAG FIND_PACKAGE_ARGUMENTS)
 
   cmake_parse_arguments(CPM_ARGS "" "${oneValueArgs}" "" ${ARGN})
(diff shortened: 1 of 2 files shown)


TheLartians:
Hey, thanks for the PR! That policy change definitely sounds like a good idea. However, I feel a bit uneasy with changing global CMake policies from a utility script like CPM, as it can be unexpected for users. E.g. this lead to different behaviour of the same scripts based on the current CMake version.

Alternatively, how about adding a note to the Readme explaining the issue and suggesting using `cmake_minimum_required(VERSION <min>...3.24)` in projects if possible?

robertmaynard:
> Hey, thanks for the PR! That policy change definitely sounds like a good idea. However, I feel a bit uneasy with changing global CMake policies from a utility script like CPM, as it can be unexpected for users. E.g. this lead to different behaviour of the same scripts based on the current CMake version.
> 
> Alternatively, how about adding a note to the Readme explaining the issue and suggesting using `cmake_minimum_required(VERSION <min>...3.24)` (which will auto-set the new policies) in projects, if possible?

I think we would need to suggest that the projects set `CMAKE_POLICY_DEFAULT_CMP0135` value instead cmake_minimum_version. That way when a nested project calls CPM it uses the correct behavior instead of the behavior being reset by the nested projects `cmake_minimum_version` call.

robertmaynard:
How about if we do this as a global CPM option? That would be an opt-in behavior change and we would document the flag as only working with CMake 3.24+

TheLartians:
> I think we would need to suggest that the projects set CMAKE_POLICY_DEFAULT_CMP0135 value instead cmake_minimum_version. That way when a nested project calls CPM it uses the correct behavior instead of the behavior being reset by the nested projects cmake_minimum_version call.

I see, that's a good point. Maybe having an option `CPM_SET_RECOMMENDED_CMAKE_POLICIES`,  that sets the policies and defaults (as available for the current CMake version) on launch, would be a good idea.

robertmaynard:
> > I think we would need to suggest that the projects set CMAKE_POLICY_DEFAULT_CMP0135 value instead cmake_minimum_version. That way when a nested project calls CPM it uses the correct behavior instead of the behavior being reset by the nested projects cmake_minimum_version call.
> 
> I see, that's a good point. Maybe having an option like `CPM_SET_RECOMMENDED_CMAKE_POLICIES`, that sets the policies and defaults (as available for the current CMake version) on launch, would be a good idea.

I have updated this PR with the introduction of `CPM_SET_RECOMMENDED_CMAKE_POLICIES`

cmake/CPM.cmake
```diff
@@ -476,6 +502,8 @@ endfunction()
 
 # Download and add a package from source
 function(CPMAddPackage)
+  cpm_set_cmake_policies()
```
TheLartians, line 505: Does it make sense to update the policies on every new package added? If we set the defaults in the beginning of the CPM script,I think it should work just as well, unless the user explicitly decides to change the default policies back.

TheLartians (commented):
Thanks for updating the PR to add the recommended policy updates!

TheLartians:
I also wonder if it would make sense to default `CPM_SET_RECOMMENDED_CMAKE_POLICIES` to `YES`, as I can imagine this being much more common than opting out, as it's hard to imagine the policy changes in question to introduce bugs. But then again I'm not completely about all possible implications and edge-cases. 

@iboB do you have any thoughts on this?

cmake/CPM.cmake
```diff
@@ -567,6 +595,13 @@ function(CPMAddPackage)
     endif()
 
     list(APPEND CPM_ARGS_UNPARSED_ARGUMENTS URL "${CPM_ARGS_URL}")
+
+    # Prefer to use the download time for timestamp, instead of the timestamp in the archive unless
+    # explicitly set by user. This allows for proper rebuilds when a projects url changes
+    if(POLICY CMP0135)
+      cmake_policy(SET CMP0135 NEW)
+      set(CMAKE_POLICY_DEFAULT_CMP0135 NEW)
+    endif()
```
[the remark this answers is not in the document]
robertmaynard: Yes it can

cmake/CPM.cmake
robertmaynard: I wanted to have the change of the policies occur when CPM was first called instead of first included. That seemed to me when people would expect the behavior of `CPM_SET_RECOMMENDED_CMAKE_POLICIES` to be activated.

I think people would presume the following would work:

```cmake
set(CPM_SET_RECOMMENDED_CMAKE_POLICIES ON)
CPMAddPackage(....)
```

robertmaynard: The upside of going with the whole file policy setting approach is we could control the scope of influence to be just the CPM functions.

Currently since I am not pushing/popping policy scopes this behavior will leak to the callers CMake code.

iboB:
I agree with @TheLartians that setting the policies to new by default is desirable. This is the future, right? :)

I also think that we should set them when CPM.cmake is included.

It should be the responsibility of the users to manually handle policies if they want to depend on the old behavor

IMO the tougher question is what to do with the policies set in `cmp_add_subdirectory`. We can just delete the code there and assume they're being handled from above (and I favor this), but it should be noted that if we allow opting out of setting the default CPM policies, old working code with options will break. If we do keep the code there, there is no point in setting them to `NEW` at the point of inclusion. Also in such case `CPMAddPackage` will set them to `NEW` in packages added with `OPTIONS` anyway regardless of the opt-out. We could also do a `cmake_policy(PUSH/POP)` there to try to save the policy configuration from the scope above, but they will be set to NEW in the package's CMakeLists anyway.

Thinking about it, I find it unpleasant that adding the exact same package with and without `OPTIONS` will lead to potentially different policy configurations in the package itself.

So... in short I would:

* set all three policies to `NEW` when including CPM.cmake **without the possibility to opt-out**
* document that these policies are set to `NEW` and advise authors that want to depend on the old behavior that they should manually take care of it in the appropriate places.

robertmaynard:
Based on the feedback it looks like everyone agrees that the policies should be set on inclusion of CPM.cmake. So I have updated the PR to do exactly that.

The policies are still guarded by `CPM_SET_RECOMMENDED_CMAKE_POLICIES`, but by default the value is set to `ON`.

TheLartians:
Thanks @iboB for the feedback and @robertmaynard for the changes! 

After thinking about it I'm now also convinced that setting the policies initially **without option to opt-out** is the way to go. Besides the points mentioned above, providing no opt-out also means less options to handle and maintain for us. If setting the policies does lead to an increase in issues, we can still add an opt-out in a future version if deemed necessary.

robertmaynard:
> 
> After thinking about it I'm now also convinced that setting the policies initially **without option to opt-out** is the way to go. Besides the points mentioned above, providing no opt-out also means less options to handle and maintain for us. If setting the policies do lead to an increase in issues, we can still add an opt-out in a future version if deemed necessary.

I have updated the PR to always enable the policies, and dropped the `CPM_SET_RECOMMENDED_CMAKE_POLICIES` option ( and docs ).

TheLartians (approved):
Looks good, thanks! Perhaps we can still add a quick note in the readme about the policies , but IMO it's also good as is.

PR #379, opened 2022-07-27
CPM_DIRECTORY always stores absolute non symlink paths

robertmaynard:
When CPM is brought into a project via downloading it will be placed in the build directory. When the build directory is a symlink the CPM_DIRECTORY value will hold the symlink value. This means that trying to run CMake via the real build directory will cause CPM to fail since it considers the current CMAKE_CURRENT_LIST_DIR to be different and will early exit.

This resolves the issue be always getting CMake to resolve symlinks before comparing paths for CPM_DIRECTORY.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -30,9 +30,9 @@ cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
 
 set(CURRENT_CPM_VERSION 1.0.0-development-version)
 
+get_filename_component(CPM_CURRENT_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}" REALPATH)
 if(CPM_DIRECTORY)
-  get_filename_component(_CPM_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}" REALPATH)
-  if(NOT CPM_DIRECTORY STREQUAL _CPM_DIRECTORY)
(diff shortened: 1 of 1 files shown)


cmake/CPM.cmake
```diff
@@ -31,7 +31,8 @@ cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
 set(CURRENT_CPM_VERSION 1.0.0-development-version)
 
 if(CPM_DIRECTORY)
-  if(NOT CPM_DIRECTORY STREQUAL CMAKE_CURRENT_LIST_DIR)
+  get_filename_component(_CPM_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}" REALPATH)
```
TheLartians, line 34: - Could we rename the variable something more descriptive like `CPM_CURRENT_DIRECTORY`? 
- Also I think the `get_filename_component` statement can be moved above the `if` statement to avoid duplication, as it exists in both branches.

TheLartians (changes requested):
Thanks the the PR! Definitely makes sense to catch this edge case.

cmake/CPM.cmake
robertmaynard: Good points, updated to do both.

TheLartians (approved):
Looks good, thanks for the changes!

PR #380, opened 2022-08-01
Add Methane Kit project to showcase table in ReadMe

egorodet:
[Methane Kit](https://github.com/MethanePowered/MethaneKit) project was recently migrated from [using Git submodules](https://github.com/egorodet/MethaneExternals) to CPM.cmake for external dependencies management, which greatly simplified dependency updates and allowed reusing the same dependencies in the derived projects without code duplication (for example [MethaneAsteroids](https://github.com/MethanePowered/MethaneAsteroids)). Please add Methane Kit to the CPM showcase table, if it's well suited for demonstration needs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,14 @@ If you know others, feel free to add them here through a PR.
         <p align="center"><b>klogg - fast advanced log explorer</b></p>
       </a>
     </td>
+    <td>
+      <a href="https://github.com/MethanePowered/MethaneKit">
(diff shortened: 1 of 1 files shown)


TheLartians (approved):
Hey, thanks for adding Methane to the showcase! Super cool project and definitely fits here. 
I might check it out myself once it supports iOS (and maybe also wasm) 😉.

PR #381, opened 2022-08-01
Use 3 column layout in showcase

TheLartians:
This gives the showcase more space, especially improving readability on mobile devices.


iboB (approved):
It does look better this way

Probably revert back to 4 columns when there are 8 or more items to showcase :grin:

PR #383, opened 2022-08-04
Document policies being set to NEW when including CPM.cmake

iboB:


TheLartians (approved):
Nice, thanks for adding the docs!

PR #386, opened 2022-08-08
Perform policy changes only in top-level CPM.cmake script

TheLartians:
An oversight from #372 on my side, but I think we should perform policy updates (or any change to the CMake state) only for the topmost CPM.cmake script as it will have the most recent policy preferences. This PR moves the policy changes below the include guard, allowing the only topmost CPM.cmake script to dictate which policies we want to enable / disable.


iboB (approved):
I agree

PR #388, opened 2022-08-11
Restore policy changes in CPMAddPackage

### LINKED ISSUE #387: Refactoring of CPM policies causes a regression in policy propagation
After deploying CPM 0.35.4 I found a regression in policy propagation.

Here is a small reproducer of the issue:

Root CMakeLists.txt
```cmake
cmake_minimum_required(VERSION 3.10.0 FATAL_ERROR)
project(cpm_reproducer)

function(cpm_init)
  include(${CMAKE_CURRENT_SOURCE_DIR}/cpm.cmake)
endfunction()
cpm_init()

add_cpm_subdir(src)
```
cpm.cmake:
```cmake
macro(set_policies )
  # the policy allows us to change options without caching
  cmake_policy(SET CMP0077 NEW)
  set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)

  # the policy allows us to change set(CACHE) without caching
  if(POLICY CMP0118)
    cmake_policy(SET CMP0118 NEW)
    set(CMAKE_POLICY_DEFAULT_CMP0118 NEW)
  endif()
endmacro()
set_policies()

function(add_cpm_subdir dir)
  # set_policies()
  add_subdirectory(${dir})
endfunction()
```

and the src/CMakeLists.txt
```cmake
message(STATUS "before src cmake_minimum_required")
cmake_policy(GET CMP0077 77_state)
cmake_policy(GET CMP0118 126_state)
message(STATUS "CMP0077 ${77_state}")
message(STATUS "CMP0118 ${126_state}")

cmake_minimum_required(VERSION 3.1.0 FATAL_ERROR)

message(STATUS "after src cmake_minimum_required")
cmake_policy(GET CMP0077 77_state)
cmake_policy(GET CMP0118 126_state)
message(STATUS "CMP0077 ${77_state}")
message(STATUS "CMP0118 ${126_state}")
```

The behavior we see is that `src/CMakeLists.txt` will reset to `3.1.0` policy settings since the variables `CMAKE_POLICY_DEFAULT_CMP0077` and `CMAKE_POLICY_DEFAULT_CMP0126` are not set in the scope the are called from.

So while CPM.cmake behaves as expected the consuming project doesn't which means that `options` and `set(CACHE)` variables will ignore the provided values we propagate.

**Proposed Solution:**

We refactor the CPM policies to set in a macro, which we call when CPM.cmake is included and also when `CPMAddPackage` is called.

robertmaynard:
Fixes #387

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -67,7 +67,7 @@ endif()
 
 set_property(GLOBAL PROPERTY CPM_INITIALIZED true)
 
-macro(cpm_set_policies )
+macro(cpm_set_policies)
   # the policy allows us to change options without caching
   cmake_policy(SET CMP0077 NEW)
   set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -67,22 +67,25 @@ endif()
 
 set_property(GLOBAL PROPERTY CPM_INITIALIZED true)
 
-# the policy allows us to change options without caching
-cmake_policy(SET CMP0077 NEW)
-set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
+macro(cpm_set_policies )
+  # the policy allows us to change options without caching
+  cmake_policy(SET CMP0077 NEW)
+  set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
 
-# the policy allows us to change set(CACHE) without caching
-if(POLICY CMP0126)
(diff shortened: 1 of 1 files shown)


TheLartians (changes requested):
Looks good to me, could you run CMake format to match the style? E.g. through

```bash
```

TheLartians (approved):

robertmaynard:
> Looks good to me, could you run CMake format to match the style? E.g. through
> 
> ```shell
> cmake -Stest/style -Bbuild/style
> cmake --build build/style --target fix-format
> ```

Updated.

Issue #389, opened 2022-08-14
Use Renovate to update versions


JohT:
I recently wrote a blog article about how to customize [Renovate](https://github.com/renovatebot/renovate) to automatically update dependency versions in [package-lock.cmake](https://github.com/cpm-cmake/CPM.cmake/wiki/Package-lock) and the download of CPM.cmake itself: 

[Keep your C++ dependencies up-to-date with Renovate & CPM](https://joht.github.io/johtizen/automation/2022/08/03/keep-your-cpp-dependencies-up-to-date.html)

I thought this might be helpful for others here...

TheLartians:
Hey this looks really neat, thanks for sharing!

JohT:
Since this isn't an open issue, I'll close it.

Issue #318, opened 2021-12-31
Using of dynamic vMajor.Minor tags


DNKpp:
Currently its rather tedious tkeeping all projects using CPM up2date. I use the ``get_cpm`` module, which works quite nice. I think it would be a great improvement, if CPM keeps additional dynamic ``vMajor.Minor`` tags, which will then be moved to the current active patch tag. These tags could then be targeted by the ``get_cpm`` module.
Image a tag ``v0.42``. Then patch  ``v0.42.1`` comes out and the ``v0.42`` tag will be reset to point exactly to the same commit as the ``v0.42.1``. But it will not move to a ``v0.43.x`` patch.

As patches usually don't break anything but patches bugs, that should be a preferable behaviour.

TheLartians:
> As patches usually don't break anything but patches bugs, that should be a preferable behaviour.

In a perfect world yes, but mistakes are easily made. Tbh I prefer strict versioning as it would guarantee reproducibility in all circumstances. Perhaps an auto-update feature (that would override the version in cmake or in a version lock file) would be nice instead.

saxbophone:
Maybe a `SEMVER` option in `CPMFindPackage()` that is an alternative to `VERSION`, and which is an explicit assertion by the user that they trust their upstream dependency follows semantic versioning and they are happy to have the `v0.4 binds to v0.4.x` behaviour that is mentioned?

TheLartians:
In general I think it's a nice idea, however should come together with proper lockfile support so that we can still commit exact versions into our repositories.

PR #394, opened 2022-08-24
Add support for conditional dependencies via INCLUDE_IF keyword.

itavero:
Still a work in progress. See #391 for more details.

To do (amongst other things probably):

- [ ] Write tests
- [ ] Update documentation

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -262,13 +262,40 @@ endfunction()
 
 # Find a package locally or fallback to CPMAddPackage
 function(CPMFindPackage)
-  set(oneValueArgs NAME VERSION GIT_TAG FIND_PACKAGE_ARGUMENTS)
+  set(oneValueArgs NAME VERSION GIT_TAG FIND_PACKAGE_ARGUMENTS INCLUDE_IF)
 
   cmake_parse_arguments(CPM_ARGS "" "${oneValueArgs}" "" ${ARGN})
 
   if(NOT DEFINED CPM_ARGS_VERSION)
     if(DEFINED CPM_ARGS_GIT_TAG)
       cpm_get_version_from_git_tag("${CPM_ARGS_GIT_TAG}" CPM_ARGS_VERSION)
+    endif()
+  endif()
+
+  if(DEFINED CPM_ARGS_INCLUDE_IF)
+
+    # Filter out INCLUDE_IF from arguments
+    cpm_remove_one_value_arg(INCLUDE_IF ARGN ${ARGN})
+
+    # If INCLUDE_IF evaluates to false, do not add the package, but do add it to the package lock
+    # and register it.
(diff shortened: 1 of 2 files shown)


cmake/CPM.cmake
```diff
@@ -272,6 +272,33 @@ function(CPMFindPackage)
     endif()
   endif()
 
+  if(DEFINED CPM_ARGS_INCLUDE_IF)
+
+    # Filter out INCLUDE_IF from arguments
+    cpm_remove_one_value_arg(INCLUDE_IF ARGN ${ARGN})
```
itavero, line 278: This function currently assumes that the user has always supplied a value for `INCLUDE_IF`.
That might be okay, but for CMake 3.15 and newer, we could also check if `INCLUDE_IF` is not in `CPM_ARGS_KEYWORDS_MISSING_VALUES` and otherwise inform this function about the fact that there is no value to be removed (see [cmake_parse_arguments docs](https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html).

iboB:
I'm against this. I'm pasting my comment from the linked discussion:

The proposed solutions would only work if all packages are listed in a central `CMakeLists`, and if this was the case, there would not be a point in having a package-lock. This central `CMakeLists` can play the same role.

But packages are listed in multiple CMakeLists files. My root can have code like:

```cmake
if(BUILD_SERVER)
    add_subdirectory(server)
endif()
if(BUILD_CLIENT)
    add_subdirectory(client)
endif()
```

In such case the dependencies of the client won't even be mentioned, if only `BUILD_SERVER` is true and vice versa. And yes, as @itavero mentioned, different build flavors may require different versions of the same package, leading to an adverse package lock.

Thinking about all edge cases and different scenarios I'm even inclined to propose ditching package-lock altogether. Has anyone made a good use of it?

iboB:
I think that package-lock should be at least rethought before proceeding with such changes

TheLartians:
Yeah I definitely hadn't given it enough thought until now. Also sharing my comment from the discussion for completeness.

> That's a really good point, package-lock definitely isn't compatible with conditional includes. Also to properly update the lock for nested dependencies we would need to somehow query any disabled packages as well making this feature much more difficult to implement than I initially anticipated.
> 
> I'm still in favour of keeping the feature, as having a package lock is crucial if we want to support automatic package updates (e.g. https://github.com/cpm-cmake/CPM.cmake/issues/318, https://github.com/cpm-cmake/CPM.cmake/issues/389), but we should make it clear that it doesn't work with conditional includes. As for INCLUDE_IF, it is still possible but would require running a separate temporary CMake process to resolve any nested dependencies when updating the package lock.
> 
> Going forward I do think the feature needs a hefty update and be used together with dynamic version tags as suggested in https://github.com/cpm-cmake/CPM.cmake/issues/318.

PR #396, opened 2022-09-02
cpm_find_package Use found package version when possible

robertmaynard:
Instead of printing the requested version, print the found version when possible. This gives better information to the users on which exact version was found locally.


iboB (approved):
Looks good

TheLartians (approved):
Thanks!

PR #397, opened 2022-09-04
Expand absolute path later in get_cpm.cmake

kisielk:
Supports expanding tildes in the environment.

### DIFF
--- a/cmake/get_cpm.cmake
+++ b/cmake/get_cpm.cmake
@@ -1,25 +1,21 @@
 set(CPM_DOWNLOAD_VERSION 1.0.0-development-version)
 
 if(CPM_SOURCE_CACHE)
-  set(CPM_DOWNLOAD_LOCATION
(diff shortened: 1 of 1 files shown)


iboB:
To make things clear: you need this because you have a tilde in an env var?

kisielk:
Yes for example in CLion I was setting the `CPM_SOURCE_CACHE` environment variable for a build with a tilde as my home directory and it was not expanded. It led to some cryptic errors. This would allow that usage and I don't think it has any other negative consequences.

iboB:
I'd say that it was CLion's fault for not expanding the tilde. Generally having literal tildes in env vars is considered bad practice. 

This indeed doesn't seem like it could break stuff.

iboB (approved):

TheLartians (approved):
Agreed, this doesn't seem like it would cause any issues.

PR #401, opened 2022-09-07
Don't warn about dirty source trees when a PATCH_COMMAND is provided

robertmaynard:
The patch command will always cause a dirty source tree, which is expected and not a cause for concern.


iboB (approved):
Makes sense

TheLartians (approved):
Agreed, thanks for the PR!

PR #413, opened 2022-09-27
Update example in readme

TheLartians:
Updates the example to showcase adding multiple dependencies and move the example to a more prominent position, as I feel it helps communicate the main ideas of CPM.cmake faster than the usage instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ CPMAddPackage("gh:nlohmann/json@3.10.5")
 CPMAddPackage("gh:catchorg/Catch2@2.5.0")
 
 # link dependencies
-target_link_libraries(main fmt nlohmann_json::nlohmann_json Catch2)
+target_link_libraries(main fmt::fmt nlohmann_json::nlohmann_json Catch2::Catch2)
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,6 +21,30 @@ For everything else, the targets can be created manually after the dependency ha
...
+
+# add dependencies
+include(cmake/CPM.cmake)
+
+CPMAddPackage("gh:fmtlib/fmt#7.1.3")
+CPMAddPackage("gh:nlohmann/json@3.10.5")
+CPMAddPackage("gh:catchorg/Catch2@2.5.0")
+
+# link dependencies
+target_link_libraries(main fmt nlohmann_json::nlohmann_json Catch2)
```
iboB, line 43: why is `nlohmann_json` namespaced and the others are not?

TheLartians: Good point, I wasn't aware that the other libraries also create namespaced targets.

```suggestion
target_link_libraries(main fmt::fmt nlohmann_json::nlohmann_json Catch2::Catch2)
```

iboB (approved):

PR #422, opened 2022-11-23
Small helper function for logging :

### LINKED ISSUE #328: CPM_INDENT is not inserted everywhere and no CPM_DESINDENT

flagarde:
This PR add an helper function only when `message(STATUS ...)` as the other are `backtracing` and adding the helper function would add on stack.

It should fix #328 #342 with a few line of code.

Default behaviour : 
Screenshot_20221123_112206
 
Possible output redefining the `cpm_message` : 
Screenshot_20221123_111315

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -28,15 +28,25 @@
 
 cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
 
+if (NOT DEFINED CPM_INDENT)
+  set(CPM_INDENT "CPM: ")
+endif ()
+
+if (NOT COMMAND cpm_message)
+  function(cpm_message)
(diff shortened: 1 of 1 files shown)


TheLartians (approved):
Thanks for the PR, looks good!

PR #426, opened 2022-11-23
get_cpm: Recover from failed download

haenkel:
If the download fails during the cmake config step an empty `CPM_<version>.cmake` file is created.
So far the script only checks the files existence, so there is no second try until the empty file is deleted manually.

This adds a check if the file is empty
and resumes the download on next config if it is.

### DIFF
--- a/cmake/get_cpm.cmake
+++ b/cmake/get_cpm.cmake
@@ -10,12 +10,23 @@ endif()
 
 # Expand relative path. This is important if the provided path contains a tilde (~)
 get_filename_component(CPM_DOWNLOAD_LOCATION ${CPM_DOWNLOAD_LOCATION} ABSOLUTE)
-if(NOT (EXISTS ${CPM_DOWNLOAD_LOCATION}))
+
+function(dl_cpm)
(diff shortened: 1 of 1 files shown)


cmake/get_cpm.cmake
```diff
@@ -10,12 +10,23 @@ endif()
 
 # Expand relative path. This is important if the provided path contains a tilde (~)
 get_filename_component(CPM_DOWNLOAD_LOCATION ${CPM_DOWNLOAD_LOCATION} ABSOLUTE)
-if(NOT (EXISTS ${CPM_DOWNLOAD_LOCATION}))
+
+function(dl_cpm)
```
TheLartians, line 14: Can we rename it to a more verbose function, e.g. `download_cpm`?

TheLartians (commented):
Thanks for the PR! Can definitely see the use of this.

cmake/get_cpm.cmake
haenkel: yes sure (did that now)

TheLartians (approved):
Thanks!

Issue #424, opened 2022-11-23
Parallel configurations of the same project can result in corrupted cpm-cache


Febbe:
When configuring the same project twice in parallel and with cpm cache enabled, it can happen, that a dependency is created with an empty / corrupted git repository. The configuration will then fail for both.

TheLartians:
Parallel configurations should now be supported in `v0.37.0`.

Issue #371, opened 2022-07-13
Failure when CPM_SOURCE_CACHE is specified with CLion


Banderi:
I'm using the latest version of CPM, freshly downloaded. Using CLion, CMake 3.24/3.22 and Windows 10.
If `CPM_SOURCE_CACHE` is set, it "fails" and the script configuration crashes, complaining that the folders already exist. (They do not. This is from a completely clean folder with no cache directories.)
This is the CMake script I have:
```cmake
set(CPM_SOURCE_CACHE .cache/CPM)
CPMAddPackage(
    GIT_TAG release-2.0.22
    GITHUB_REPOSITORY libsdl-org/SDL
)
```
And this is the output:
```
CPM_SOURCE_CACHE: .cache/CPM
-- CPM: adding package SDL@ (release-2.0.22 to E:/Git/Antimony-III/.cache/CPM/sdl/22376c56c7faadb422628d46bf5157d6eb41d577)
[1/9] Creating directories for 'sdl-populate'
[1/9] Performing download step (git clone) for 'sdl-populate'
fatal: destination path '22376c56c7faadb422628d46bf5157d6eb41d577' already exists and is not an empty directory.
fatal: destination path '22376c56c7faadb422628d46bf5157d6eb41d577' already exists and is not an empty directory.
fatal: destination path '22376c56c7faadb422628d46bf5157d6eb41d577' already exists and is not an empty directory.
-- Had to git clone more than once:
          3 times.
CMake Error at E:/Git/Antimony-III/cmake-build-release/_deps/sdl-subbuild/sdl-populate-prefix/tmp/sdl-populate-gitclone.cmake:31 (message):
  Failed to clone repository: 'https://github.com/libsdl-org/SDL.git'
```

Upon reloading the CMake config, it also either "fails" to remove the previous cache folders -
```
Error removing directory "E:/Git/Antimony-III/.cache/CPM/sdl2/22376c56c7faadb422628d46bf5157d6eb41d577".
CMake Error at E:/Git/Antimony-III/cmake-build-release/_deps/sdl2-subbuild/sdl2-populate-prefix/tmp/sdl2-populate-gitclone.cmake:12 (message):
  Failed to remove directory:
  'E:/Git/Antimony-III/.cache/CPM/sdl2/22376c56c7faadb422628d46bf5157d6eb41d577'
```
- or complains that the cache is dirty:
```
CMake Warning at cmake/CPM.cmake:679 (message):
  Cache for SDL2
[...]

Issue #293, opened 2021-10-07
Multiple parallel configure with Colcon and CPM: not supported


huand:
Hello,
Im not too sure this is the cause of the problem, but when using colcon (a build tool for ros2 packages), I think colcon starts all leaf dependencies config and build in "parallel", if there are multiple duplicate dependencies among those parallel build, CPM fails. Example, where MyProject1 is used by several leaf projects  and downloaded by cpm:
````
[ 11%] Creating directories for 'MyProject1'
[ 22%] Performing download step (git clone) for 'MyProject1'
Cloning into 'fc4e6fc51c1494235cb68e60c5b7fb54c701a612'...
fatal: shallow file has changed since we read it
fatal: destination path 'fc4e6fc51c1494235cb68e60c5b7fb54c701a612' already exists and is not an empty directory.
fatal: destination path 'fc4e6fc51c1494235cb68e60c5b7fb54c701a612' already exists and is not an empty directory.
-- Had to git clone more than once:
          3 times.
````
however if i start the process again, then it works.
would it be possible for CPM to check if another CPM is already active for some deps and if another cpm process wants the same deps to just wait for the other to finish instead of failing? I understand this is probably a niche case for build system like Colcon.

egorodet:
I'm experiencing a similar issue in [CLion](https://www.jetbrains.com/clion/) IDE, which allows to work with multiple CMake configurations and runs `cmake configure` for active configurations in parallel. In such case it became impossible to use `CPM_SOURCE_CACHE` because of update collisions, so I had to fallback to separate source locations, which seems to be slowly working.

TheLartians:
Is this fixed in v0.37.0 with #427?

egorodet:
> Is this fixed in v0.37.0 with #427?

@TheLartians I've just tested v0.37 with multiple CMake configurations initializing dependencies in parallel in CLion and it worked just fine without any errors! Thanks a lot for fixing this painful issue!

TheLartians:
Great, thanks for testing! I close the issue for now as I assume this directly relates to the Colcon error as well.

PR #427, opened 2022-11-27
Add file locking to support parallel runs.

PercentBoat4164:
This should fix some issues pointed out by other users. Specifically #424, #371. and #293. Only CPMAddPackage() was changed. Changes should be made anywhere else where the cache is accessed.

Thanks to @Banderi for providing the idea for this solution.


PercentBoat4164:
My fork's Actions have returned this error while executing `build_all.py`.

```C
error while running 'cmake --build build/doctest -- -j1':
   In file included from /home/runner/work/CPM.cmake/CPM.cmake/examples/doctest/main.cpp:3:
  /home/runner/work/CPM.cmake/CPM.cmake/build/doctest/_deps/doctest-src/doctest/doctest.h:4299:47: error: size of array ‘altStackMem’ is not an integral constant-expression
   4299 |         static char             altStackMem[4 * SIGSTKSZ];
        |                                               ^
  gmake[2]: *** [CMakeFiles/CPMExampleDoctest.dir/build.make:76: CMakeFiles/CPMExampleDoctest.dir/main.cpp.o] Error 1
  gmake[1]: *** [CMakeFiles/Makefile2:132: CMakeFiles/CPMExampleDoctest.dir/all] Error 2
  gmake: *** [Makefile:146: all] Error 2
```

However, I am unable to look into this problem because my local development environment is unable to build even default CPM's master branch or the latest release. Building the `test-verbose` target results in test 1 and 5 failing. Both fail on line 34 of their respective `CPM.cmake` with the following error: (5 provided as example).

```CMake
CMake Error at /home/<user>/Documents/Programming/C++/CPM.cmake/test/cmake-build-debug/fetchcontent_dependency/_deps/fibonacci-src/cmake/CPM.cmake:34 (if):
5:   if given arguments:
5: 
5:     "NOT" "/home/<user>/Documents/Programming/C++/CPM.cmake/cmake" "MATCHES" "/home/<user>/Documents/Programming/C++/CPM.cmake/test/cmake-build-debug/fetchcontent_dependency/_deps/fibonacci-src/cmake"
5: 
5:   Regular expression
5:   "/home/<user>/Documents/Programming/C++/CPM.cmake/test/cmake-build-debug/fetchcontent_dependency/_deps/fibonacci-src/cmake"
5:   cannot compile
5: Call Stack (most recent call first):
5:   /home/<user>/Documents/Programming/C++/CPM.cmake/test/cmake-build-debug/fetchcontent_dependency/_deps/fibonacci-src/CMakeLists.txt:10 (include)
```

I would like to fix this problem, but I don't know where to being. It seems as if these `CPM.cmake` files are not the same as the one found on the master branch, lastest release, or my fork, (all of which do not have this error.) so I don't know where they are coming from. Is it worth fixing my development environment or is the first problem an easy fix?

For reference, line 34 of this `CPM.cmake` looks like this:
```Cmake
if(NOT ${CPM_DIRECTORY} MATCHES ${CMAKE_CURRENT_LIST_DIR})
```
Looking through the rest of the file makes it obvious that it is completely different from any of the other current ones. Even the copyright comment is different saying `Copyright (c) 2019 Lars Melchior` as opposed the current `Copyright (c) 2019-2022 Lars Melchior and contributors`.

As for the first problem, it seems unrelated to my change, but I suppose it must be because it must not be in the master branch. I am just a little confused and would appreciate guidance.

oddko:
This should only be done when `CPM_SOURCE_CACHE` is used IMO (theres no issue doing parallel builds when modules are cloned in separate build directories), and perhaps lock files could be placed inside the subdirectory (CPM_SOURCE_CACHE/packagename) dedicated to storing the different versions of said packages which could improve parallelism a bit.

PercentBoat4164:
I did implement only locking the download directory but that caused half of the integration tests to fail. I do not know enough about how those work or what they check for to be able to debug them, so I reverted back to locking the `CPM_SOURCE_CACHE` directory. This fixes all of the tests, but the examples still fail to build both in my local development environment and in GitHub's Actions runner.

PercentBoat4164:
What! Why does it do anything different here? All of the environments behave differently! I probably just have things setup wrong. Thanks for looking into this. Should I try to make locking just the download directory work, or will this be merged as-is when someone gets a chance to review it?

TheLartians:
TBH I don't have much experience with CMake locks, so I don't know what edge cases to look out for (e.g. will things be unlocked if CMake errors/crashes etc?) in this PR. @iboB do you have more experience and could take over?

PercentBoat4164:
If CMake crashes then the locks will be released. The locks only exist within the scope of the process according to [CMake's documentation](https://cmake.org/cmake/help/latest/command/file.html#lock).
> File will be locked for scope defined by GUARD option (default value is PROCESS). RELEASE option can be used to unlock file explicitly.

oddko:
It uses file locking so it cannot possibly outlive the CMake process. It's fairly reliable (I have come to use it to secure the auto generation of in-source files).

As for locking more locally, I meant the directory right above the download one (which is module specific but not version specific). Obviously generating it in the download directory itself will lead to a dirty git directory which can cause troubles as CPM does not attempt a checkout on dirty git repos. Keep in mind the cmake.lock files are not cleaned up after the fact.

PercentBoat4164:
OK. I misunderstood your suggestion the first time. These commits should move the locked file into the per-package directory and remove it when it is no longer needed.

PercentBoat4164:
What do I need to do to keep this moving forward? Should I merge the recent additions to CPM into my fork?

PercentBoat4164:
OK. I may have to pass this on to someone else. I think I have a decent start here but I cannot debug this. None of my environments behave like the one here does. The only time I am getting meaningful feedback is when the workflows are run here, and that is not conducive to debugging. Furthermore, I don't know anything about the error that is being raised here or why it did not appear before, or why it did appear before on my fork's workflow.
To be clear, I would happily continue to polish this pull request to perfection if someone could inform me as to the proper way to set up the workflows, or optimally, my local environment such that they match the ones here, or even if someone were to provide documentation to the same effect.
Sorry about this, but it is a little frustrating to be limited in this way.

TheLartians:
@PercentBoat4164 it seems that the GitHub actions runner has been updated to use a new version of compiler tools that are no longer compatible with some of the example projects being built in the test suite. This should be soon resolved in #430 which should hopefully solve the issues you are seeing in the workflows.

PercentBoat4164:
Oh! That is amazing to hear. Thank you. I am glad that I am not going insane. Is there anything that I need to do to prepare for this change?

CraigHutchinson:
@PercentBoat4164 I like this but I cannot see any additional unit-test to check the added behavior? This may be tricky but as CPM is being used by many more developers we need to ensure features don't get broken inadvertently by the next person who comes along 👍 

As an additional 'idea' all packages are currently satisfied sequentially and parallelism is inherently lost. A future enhancement could be to 'somehow' check the lock with `TIMEOUT 0` and process the subsequent dependencies. However I cannot see how this would work without some new/extra fence/barrier function to await-all dependencies being unlocked 🤔

PercentBoat4164:
Hmm... OK. The unit-tests are written in Ruby. I don't know Ruby. I can learn it though. I will come up with a unit-test to go along with this pull request's feature.

 To be clear, actually downloading two packages at once was never part my plan. I simply wanted to avoid errors that come up when I use CPM with CLion's multi-configuration CMake execution. I could imagine that this could be useful however. I too do not see a way to do that without a fence function. Perhaps a new `CPMFinish` function could be defined?

```CMake
# Project set up...

CPMAddPackage(...)
CPMAddPackage(...)
CPMFinish()
# All packages added before this point are guaranteed to be available at this point.

# Use packages...
```

Perhaps a better name is in order, but does that calling pattern fit well with what already exists in CPM?

oddko:
Won't really work as ultimately the clash is due to parallel git executions, not cmake parallel executions per se.

PercentBoat4164:
What is happening to cause multiple executions of git on different processes to interfere with each other? Furthermore, is it worth trying to find a way around it?

oddko:
Because that's what FetchContent does, it does a call to git clone (and possibly checkout if you only change the tag that is used). Doing several runs at the same time on the same repository will throw git under the bus, and that'd be the whole point of adding a lock: making only one git command run at the same time.

cmake/CPM.cmake
```diff
@@ -790,6 +798,12 @@ function(CPMAddPackage)
     cpm_get_fetch_properties("${CPM_ARGS_NAME}")
   endif()
 
+  if(CPM_HAS_CACHE_LOCK)
+    file(LOCK ${download_directory}/../cmake.lock RELEASE)
+    file(REMOVE ${download_directory}/../cmake.lock)
+    set(CPM_HAS_CACHE_LOCK OFF)
+  endif()
```
TheLartians, line 805: Would this break when a dependency itself uses CPM to download additional dependencies? 

Afaict inside the inner `CPMAddPackage` call, `CPM_HAS_CACHE_LOCK` would be inherited from the outer scope, resulting in them not creating a lock, but releasing and removing any locks relative to their download directory.

Maybe it could be placed before the `cpm_add_subdirectory` call?

cmake/CPM.cmake
```diff
@@ -156,6 +156,8 @@ set(CPM_SOURCE_CACHE
     CACHE PATH "Directory to download CPM dependencies"
 )
 
+set(CPM_HAS_CACHE_LOCK OFF)
+
```
TheLartians, line 160: I think this has no implications, as it sets a local variable that is removed once the CPM.cmake file has finished evaluating. Perhaps it can be removed?

TheLartians (commented):
Hey thanks for the patience. I've finally merged the updated example versions into master and have updated your PR. Now the examples build as expected on GitHub workflows!

About testing this feature, I agree that having a unit test would be preferable, if it is easily possible. As said before I'm not really sure about the implications of this feature but as long as existing builds don't break and a deadlock likely won't occur I'm happy with moving forward.

cmake/CPM.cmake
PercentBoat4164: There are two `cpm_add_subdirectory` calls in the area between the file locking and unlocking. Which one are you thinking it should be before? If it goes before both of them, then there are calls to `cpm_fetch_package` that would be unprotected.

cmake/CPM.cmake
PercentBoat4164: Yes this can be removed.

cmake/CPM.cmake
PercentBoat4164: I took a stab at solving this problem by changing `CPM_HAS_CACHE_LOCK` to `${CPM_ARGS_NAME}_CPM_HAS_CACHE_LOCK`. This should create a new variable for each new package, preventing the variable that controls protection from being the same one that is used in the next scope.

PercentBoat4164:
I am trying to write a unit test in CMake. My particular feature requires multiple processes to test. In CMake I could not find a way to do that. I see that you already use Python for testing examples. Would it be acceptable to have a CMake script call a Python script that spawns the processes and performs the tests?

cmake/CPM.cmake
TheLartians: Understood, I believe that should work as well. Thanks for the changes!

TheLartians (dismissed):
Quite happy with the changes now, and the test does show behaviour that wasn't possible before. Thanks for implementing the feature!

TheLartians:
PS: I've gone ahead and converted your test case to an integration test for better consistency with the new test structure. For future reference, you can find more information on writing tests [here](https://github.com/cpm-cmake/CPM.cmake/tree/master/test/integration#writing-tests).

cmake/CPM.cmake
```diff
@@ -712,6 +712,12 @@ function(CPMAddPackage)
     # relative paths.
     get_filename_component(download_directory ${download_directory} ABSOLUTE)
     list(APPEND CPM_ARGS_UNPARSED_ARGUMENTS SOURCE_DIR ${download_directory})
+
+    if(NOT ${CPM_ARGS_NAME}_CPM_HAS_CACHE_LOCK)
```
oddko, line 716: Isn't this (local) variable useless since you are always going to lock and release the file anyways ?
Besides this doesnt need to be done when clones are operated into different directories (i.e. when `CPM_SOURCE_CACHE` is not used, which could be a somewhat decent condition to check ?).

cmake/CPM.cmake
```diff
@@ -790,6 +796,12 @@ function(CPMAddPackage)
     cpm_get_fetch_properties("${CPM_ARGS_NAME}")
   endif()
 
+  if(${CPM_ARGS_NAME}_CPM_HAS_CACHE_LOCK)
+    file(LOCK ${download_directory}/../cmake.lock RELEASE)
+    file(REMOVE ${download_directory}/../cmake.lock)
```
oddko, line 801: I'm not sure if it's a good idea to delete the file. Picture the (unlikely but plausible) case where right after this instance of CMake releases the lock, the other one running in parallel acquires it before yours has time to delete the file. Removing the file will fail for permission issues and this instance of CMake will terminate with an error.

PercentBoat4164: It is not too far from unlikely as all other running CMake processes are trying to get a lock on the file while this part of the code is being executed. The reason that I chose to delete the file is twofold. First, I wanted to clean up the lock files, and this is the easiest way to do it. Secondly, simply unlocking the files and not deleting them did not work on Windows. Perhaps only deleting the files is a good way to handle this?

cmake/CPM.cmake
PercentBoat4164: Yes. It looks like these variables are useless. The reason it exists is that in the past I had come across an error that occurred because the point of execution entered the locked block then left it and went back into it without going past the unlocking line. This would cause an error along the lines of `file already locked!` to occur. The local variable was used to solve that problem. I just tried it again without the `${CPM_ARGS_NAME}_CPM_HAS_CACHE_LOCK` and it succeeded, so I will happily remove the checks.

As for the `CPM_SOURCE_CACHE` check, a while ago I moved all of my changes to within a block that checked for this. see [line 700](https://github.com/PercentBoat4164/CPM.cmake/blob/a5a33b8141b8e8954f040c13bb22bd49f7aa25ea/cmake/CPM.cmake#L700). If this is not a good place for it, I can move it elsewhere and perform my own checks.

cmake/CPM.cmake
oddko: Yes but that's the whole point of using a file as a lock, so long as a cmake process can't open it (because it already is by another process), it'll be blocked in its execution, hence my remark that attempting to remove it while it gets opened by another cmake process as part of the locking process is risky imo though by the time the other cmake process wakes up you'll probably be far down the function already. I don't think you should sweat it about removing the `cmake.lock` files, given where they are located they are not bothersome at all.

So a simple lock/release combo should do the job mighty fine.

PercentBoat4164: OK. So I implemented the unlock only, but because the file still exists, the cache.cmake unit test fails because it finds two versions of fibonacci. It mistakes cmake.lock for a version. Should that be changed to ignore any non-directory entries, or should I make an effort to clean up the cmake.lock file?

TheLartians: I think keeping the lock files is fine as long as they're created only inside the CPM cache directory.
As for the test, would it be possible to glob all files not ending in `.lock`? Otherwise excluding directories is fine, though perhaps put the logic in an extra function e.g. `assertCacheDirectoryCount(<name> <count>)` to avoid repeating yourself.

iboB (approved):
I'm not an expert on how CMake handles file locking, but from what I've read. This commit looks good.

Kudos for making an integration test. 

I pushed some changes to how the requested integration test lib change is handled. I think its better to be explicit in such scenarios. Also the new arg `name` can be useful in other scenarios

iboB (dismissed):

cmake/CPM.cmake
PercentBoat4164: According to the [`file(GLOB ...)`](https://cmake.org/cmake/help/latest/command/file.html#glob) documentation for CMake, there is no way to exclude text. Therefore, I am making a function to replace this logic. I am wondering about the repetition on [line 52](https://github.com/PercentBoat4164/CPM.cmake/blob/master/test/unit/cache.cmake#L52). The variable names are changed, but otherwise I see no difference. Is it really needed?

test/unit/cache.cmake
```diff
@@ -51,7 +51,7 @@ execute_process(
 assert_equal(${ret} "0")
 assert_exists("${CPM_SOURCE_CACHE_DIR}/fibonacci")
 
-assertCacheDirectoryCount("${CPM_SOURCE_CACHE_DIR}/fibonacci/*" 1)
+assertcachedirectorycount("${CPM_SOURCE_CACHE_DIR}/fibonacci/*" 1)
```
iboB, line 54: just rename the function as `snake_case`

iboB (approved):

PercentBoat4164:
Oops. Sorry.

TheLartians (approved):
Looks great, thanks again for implementing this feature and for your patience and changes!

TheLartians:
This feature is now released in `v0.37.0`. 🎉

PR #430, opened 2022-12-17
Update example dependency versions

TheLartians:
This PR updates versions of dependencies used in examples for better compatibility with new hardware and CMake versions.

## Motivation

It seems that the GitHub runner image `ubuntu-latest` was updated to use recent versions C++ build tools, resulting in some errors from various dependencies. These seem to already be addressed in more recent versions so we can simply update the dependencies to be compatible with the newest GitHub runners.

Might fix issues currently blocking  #427 from continuing.


TheLartians:
@iboB, as this is blocking other contributions, I'll go ahead and merge without review.

PR #434, opened 2023-01-09
Add JNGL to "Built with CPM.cmake"

jhasse:
I'm using CPM to get ogg, vorbis, freetype, OpenAL Soft and webp :) Improved configure times compared to FetchContent due to the source cache significantly and I can reconfigure offline.


TheLartians (approved):
Awesome, thanks for sharing. Looks very cool!

Issue #377, opened 2022-07-23
heads up: CMake 3.25 `SYSTEM` option for FetchContent


david-fong:
[Here's the motivating issue in CMake's issue tracker](https://gitlab.kitware.com/cmake/cmake/-/issues/22401). `FetchContent` uses `ExternalProject_Add`, which uses `add_subdirectory`, which is unlike importing via `find_package` in that the targets that get added will not be `IMPORTED`, and will therefore have their `target_include_directories` not marked as `SYSTEM`, which means compilers will check those third party headers to warn for things that the user only wanted warnings for for their own project's code (not for third party header code).

[Here's the ticket tracking the solution that was proposed and is now being implemented](https://gitlab.kitware.com/cmake/cmake/-/issues/18040#note_1197146). In summary,

> 1. Add a SYSTEM property for targets
> 2. ...
> 3. As a helper based on the `SYSTEM` property, implement `add_subdirectory(.. SYSTEM)` and `FetchContent_Declare(.. SYSTEM BOOL)`

At the time of this writing, these changes are not merged in yet, and I'm not sure, but I don't think CMake 3.25 will be released until November, so it's still quite early to be mentioning this. I thought it would be nice to just put it on the radar so it doesn't go undetected.

I'm guessing there will be some policy variable that can be checked to see if the CPM user's CMake version supports these new arguments, and if so, to add a similar parameter/option to `CPMAddPackage`.

TheLartians:
Thanks for the heads up! It looks like CMake's support for external libraries is slowly improving! 

Perhaps it would be a good idea to add a document with CMake policies that impact CPM's behaviour and discuss how to handle them. Another `FetchContent` related policy change is handled in #372 .

david-fong:
quoting from a related CMake ticket ([here](https://gitlab.kitware.com/cmake/cmake/-/issues/18040#note_1220379)):

> If something relies on a feature that was introduced in a particular CMake version, then a conditional of the form `if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.xx")` would be the simplest way to check for the availability of that feature. That will only be reliable once a feature makes its way into an official release though.

jmigual:
CMake 3.25 just released the `SYSTEM` property: https://cmake.org/cmake/help/v3.25/release/3.25.html#properties

Bktero:
Hello there
Any status on this feature?
It would be highly appreciated :)

ClausKlein:
[...]

PR #441, opened 2023-02-11
Use SYSTEM option for FetchContent and add_subdirectory

ClausKlein:
first try to fix https://github.com/cpm-cmake/CPM.cmake/issues/377


ClausKlein:
@iboB @TheLartians May it possible to  review this?

I have tested it with my project: https://github.com/ClausKlein/ModernCmakeStarter/blob/develop/cmake/CPM.cmake

TheLartians:
Hey @ClausKlein thanks for making this PR and sorry for not noticing this earlier (I'm going through my notifications chronologically). In the meanwhile I've already setup a very similar PR at https://github.com/cpm-cmake/CPM.cmake/pull/449 as a response to #377, which would supersede this when merged.

TheLartians (approved):

PR #445, opened 2023-02-15
Allow defining custom prefixes

papperlapapp:
In addition to the predefined prefixes 'gh', 'gl' and 'bb', allow defining your own prefixes for, e.g. internal hosts.


TheLartians:
Hey thanks for the PR! Changes look good to me though I'd like an additional opinion from @iboB as he authored the original feature. Could you also format the code using according to the [style guide](https://github.com/cpm-cmake/CPM.cmake/blob/master/CONTRIBUTING.md#styleguides)?

cmake/CPM.cmake
```diff
@@ -353,19 +354,32 @@ function(cpm_parse_add_package_single_arg arg outArgs)
     elseif(scheme STREQUAL "bb")
       set(out "BITBUCKET_REPOSITORY;${uri}")
       set(packageType "git")
+    elseif(DEFINED CPM_CUSTOM_PREFIXES)
+      foreach(prefix ${CPM_CUSTOM_PREFIXES})
+        if("${prefix}" MATCHES "([^:]+):(.+)")
+          if(scheme STREQUAL "${CMAKE_MATCH_1}")
+            set(out "GIT_REPOSITORY;${uri};CUSTOM_REPOSITORY;${CMAKE_MATCH_2}")
```
iboB, line 361: why introduce `CUSTOM_REPOSITORY` at all? It seems somewhat redundant to me

Why not concat the string here: `"GIT_REPOSITORY;${CMAKE_MATCH_2}/${uri}"`?

iboB (commented):
I have several problem with this as it seems kinda incomplete. I'm not sure whether we want to pursue a more "complete" feature with this.

1. I'm not sure I like the name "prefix" in this case. Maybe schemes?
2. As I said, I don't like the addition of `CUSTOM_REPOSITORY`. The entire uri is contained in the long-form anyway. Adding `CUSTOM_REPOSITORY` to split in in parts seems redundant and unneeded.
3. `CPM_CUSTOM_PREFIXES` seems to imply git. This is easily fixable as we have tools to detect the type of the repo based on the uri. Alternatively it should be `CPM_CUSTOM_GIT_REPO_PREFIXES`
4. These are prefixes only. So if I wanted to add my custom GitLab instance I would have to manually add .git at the end of the uris, like:

```cmake
CPMAddPackage(my:foo/bar.git@1.2.3) # :(
CPMAddPackage(gh:foo/bar@1.2.3) # :)
```

I can't think of a good way to fix this. Something like `CPM_CUSTOM_SUFFIXES` perhaps?...

iboB:
Here's how I would do it:

```cmake
CPMAddCustomScheme(
    c1 # scheme name
    PREFIX https://c1.example.com/ # optional
    SUFFIX .git # optional
    TYPE GIT_REPOSITORY # optional. if not specified tries to infer the type from the final uri
)
```

If this is implemented, the existing default schemes can be added to the same underlying lists that this is using.

It can even allow overriding of existing schemes and some of the functionality requested in #333

cmake/CPM.cmake
papperlapapp: Because I didn't knew it better. This was the first time I fiddled with CMake functions.

papperlapapp:
> Here's how I would do it:
> 
> ```cmake
> CPMAddCustomScheme(
>     c1 # scheme name
>     PREFIX https://c1.example.com/ # optional
>     SUFFIX .git # optional
>     TYPE GIT_REPOSITORY # optional. if not specified, tries to infer the type from the composed uri
> )
> 
> # ...
> 
> CPMAddPackage(c1:foo/bar@1.2.3) # :)
> ```
> 
> If this is implemented, the existing default schemes can be added to the same underlying lists that this is using.
> 
> It can even allow overriding of existing schemes and some of the functionality requested in #333

This looks definitely better than my attempt. I already had the feeling that a generic way of implementing this should somehow be possible.

Time for diving deeper into the CMake documentation ...

Shall I close the PR and start anew, or just keep it open and work on it (will definitely take its time)?

iboB:
Awesome! It's great that you want to invest more in this

> Shall I close the PR and start anew, or just keep it open and work on it (will definitely take its time)?

I have no special opinion on the matter. However you prefer

ClausKlein:
Would this MR https://github.com/cpm-cmake/CPM.cmake/pull/528 be a the simpler solution for your problem?

k-channon-PA:
I just implemented a similar thing to this... and then checked the open PRs only to find this one!

https://github.com/cpm-cmake/CPM.cmake/commit/a6cc6b9096ad0687119a781997be504326058a72

This feature would really help me out. I'm happy to help get this PR completed if that'a any use to you guys :)

PR #450, opened 2023-02-20
Improve Multi-threaded Performance.

### LINKED ISSUE #419: How to skip `cpm_check_git_working_dir_is_clean`
Hi, when benchmarking cmake run with `cmake --profiling-format=google-trace --profiling-output=perf.json ..` I find that some huge repo takes couple of hundred of milliseconds to run  `cpm_check_git_working_dir_is_clean`.
I would like to be able to skip this check.

I see in source code:
https://github.com/cpm-cmake/CPM.cmake/blob/03705fcce165c7eb2680e4c7d8a272cfb0a1061a/cmake/CPM.cmake#L715-L721

It seems to be possible to skip it by adding `PATCH_COMMAND`, but this is per target. Is there any other way I've missed to disable the check? Ideally some `-DCPM_CHECK_GIT_WORKING_DIR_IS_CLEAN=OFF` that I can apply everywhere.
I would be happy to provide a PR, but I guess some discussion on how to do things is therefore required.

Here is also an example of benchmark:
image

As you can see, it's not lots of time, but when there is 20-30 projects, it can takes couple of seconds.

Have a nice day.

PercentBoat4164:
&nbsp;&nbsp;&nbsp;&nbsp;The idea of this pull request is to improve the performance of CPM in multi-threaded conditions by removing the `cpm_add_subdirectory()` calls from the scope of the lock.

&nbsp;&nbsp;&nbsp;&nbsp;Before this change, each instance of CMake must wait for each instance that gets hold of the lock before it to completely finish checking if the cache is clean, then adding the subdirectory. This can become a very long process for even moderately sized libraries.

&nbsp;&nbsp;&nbsp;&nbsp;To solve this problem, I propose moving the `file(LOCK ... RELEASE)` command to just after the fetch has taken place, and adding another one right before checking if the cache is clean. This will allow the first process to stop all others during the download, then when it is unlocked, all processes can simultaneously check the cache (still only done by processes that did not perform the download) and add the new subdirectory.

&nbsp;&nbsp;&nbsp;&nbsp;This will help alleviate, but will not fix #419.


iboB (approved):
This looks good to me

Issue #455, opened 2023-02-23
CPMAddPackage(NAME greeter SOURCE_DIR ..) should not call find_package()!


ClausKlein:
This `ModernCppStarter` pattern does not work if `CPM_USE_LOCAL_PACKAGES="YES"` is used and the library already installed:

```cmake
CPMAddPackage(NAME greeter SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/..)
```

```bash
cd all && cmake --preset default --debug-find-pkg=greeter
Running with debug output on for the 'find' commands for package(s) greeter.
Preset CMake variables:

  BUILD_SHARED_LIBS="YES"
  CMAKE_BUILD_TYPE="Debug"
  CMAKE_CXX_STANDARD="20"
  CMAKE_DEBUG_POSTFIX="D"
  CMAKE_INSTALL_PREFIX:PATH="/Users/clausklein/Workspace/cpp/ModernCmakeStarter/stagedir"
  CMAKE_PREFIX_PATH:STRING="/Users/clausklein/Workspace/cpp/ModernCmakeStarter/stagedir"
  CPM_USE_LOCAL_PACKAGES="YES"
  ENABLE_TEST_COVERAGE="YES"
  OPTION_ENABLE_CLANG_TIDY="YES"
  OPTION_ENABLE_UNITY="NO"

Preset environment variables:

  CPM_USE_LOCAL_PACKAGES="YES"
  PATH="/Users/clausklein/.local/bin:/usr/local/opt/net-snmp/bin:/usr/local/opt/coreutils/libexec/gnubin:/usr/local/libexec/bin:/usr/local/opt/curl/bin:/usr/local/opt/libxml2/bin:/usr/local/opt/libxslt/bin:/usr/local/opt/expat/bin:/usr/local/opt/llvm/bin:/usr/local/opt/net-snmp/bin:/Users/clausklein/Library/Python/3.9/bin:/usr/local/opt/python/libexec/bin:/usr/local/opt/sqlite/bin:/usr/local/bin:/usr/local/sbin:/Users/clausklein/scripts:/Users/clausklein/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/texbin:/usr/local/MacGPG2/bin:/opt/X11/bin:/Library/Apple/usr/bin"

[...]

PR #456, opened 2023-02-23
SOURCE_DIR option disable call to find_package()

ClausKlein:
this fix https://github.com/cpm-cmake/CPM.cmake/issues/455

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,9 @@ This can be controlled on a per package basis with the `CPM_DOWNLOAD_<dependency
 ### CPM_USE_LOCAL_PACKAGES
 
 CPM can be configured to use `find_package` to search for locally installed dependencies first by setting the CMake option `CPM_USE_LOCAL_PACKAGES`.
+
+If a packages should always be added the defined way you should set the additional parameter FORCE YES to disable the use of `find_package`
+
 If the option `CPM_LOCAL_PACKAGES_ONLY` is set, CPM will emit an error if the dependency is not found locally.
 These options can also be set as environmental variables.
 
--- a/README.md
+++ b/README.md
@@ -200,12 +200,12 @@ This can be controlled on a per package basis with the `CPM_DOWNLOAD_<dependency
 
 CPM can be configured to use `find_package` to search for locally installed dependencies first by setting the CMake option `CPM_USE_LOCAL_PACKAGES`.
 
-If a packages should always be added the defined way you should set the additional parameter FORCE YES to disable the use of `find_package`
-
 If the option `CPM_LOCAL_PACKAGES_ONLY` is set, CPM will emit an error if the dependency is not found locally.
 These options can also be set as environmental variables.
 
 In the case that `find_package` requires additional arguments, the parameter `FIND_PACKAGE_ARGUMENTS` may be specified in the `CPMAddPackage` call. The value of this parameter will be forwarded to `find_package`.
+
+Note that this does not apply to dependencies that have been defined with a truthy `FORCE` parameter. These will be added as defined.
 
 ### CPM_USE_NAMED_CACHE_DIRECTORIES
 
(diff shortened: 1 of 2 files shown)


cmake/CPM.cmake
```diff
@@ -648,19 +648,21 @@ function(CPMAddPackage)
     return()
   endif()
 
-  if(CPM_USE_LOCAL_PACKAGES OR CPM_LOCAL_PACKAGES_ONLY)
-    cpm_find_package(${CPM_ARGS_NAME} "${CPM_ARGS_VERSION}" ${CPM_ARGS_FIND_PACKAGE_ARGUMENTS})
+  if(NOT DEFINED CPM_ARGS_SOURCE_DIR)
```
TheLartians, line 651:
```suggestion
  if(NOT CPM_ARGS_FORCE)
```

TheLartians (commented):
Interesting edge case, tbh I think using the local package should be the expected behaviour when `CPM_USE_LOCAL_PACKAGES` or `CPM_LOCAL_PACKAGES_ONLY` is defined. How about instead instead using the existing `FORCE` argument to disable env based overrides?

Then packages that should **always** be added the defined way can set the additional parameter `FORCE YES`.

cmake/CPM.cmake
ClausKlein: Neither `FORCE` nor `SOURCE_DIR` are documented?

And the `SOURCE_DIR` is use at `ModernCppStarter`!

ClausKlein:
> Interesting edge case, tbh I think using the local package should be the expected behaviour when `CPM_USE_LOCAL_PACKAGES` or `CPM_LOCAL_PACKAGES_ONLY` is defined. How about instead instead using the existing `FORCE` argument to disable env based overrides?
> 
> Then packages that should **always** be added the defined way can set the additional parameter `FORCE YES`.

Yes, may be a clear Solution, but incompatible to the [SOURCE_DIR option use at ModernCppStarter](https://github.com/TheLartians/ModernCppStarter/blob/84ed75ca7db28672a29d865382389ad3e0f03f8a/standalone/CMakeLists.txt#L19)

cmake/CPM.cmake
TheLartians: > Neither `FORCE` nor `SOURCE_DIR` are documented yet?

`SOURCE_DIR` is an origin parameter of [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html#id1), so documentation can be found there. `FORCE` we should definitely document though.

TheLartians:
> Yes, may be a clear Solution, but incompatible to the [SOURCE_DIR option used at ModernCppStarter](https://github.com/TheLartians/ModernCppStarter/blob/84ed75ca7db28672a29d865382389ad3e0f03f8a/standalone/CMakeLists.txt#L19)

Yeah it seems that the starter currently doesn't support `CPM_USE_LOCAL_PACKAGES` or `CPM_LOCAL_PACKAGES_ONLY`. I think adding the `FORCE` option there would make sense.

ClausKlein:
Does this changes have a site effect to `CPMFindPackage()`, I think yes!

README.md
```diff
@@ -199,6 +199,9 @@ This can be controlled on a per package basis with the `CPM_DOWNLOAD_<dependency
 ### CPM_USE_LOCAL_PACKAGES
 
 CPM can be configured to use `find_package` to search for locally installed dependencies first by setting the CMake option `CPM_USE_LOCAL_PACKAGES`.
+
+If a packages should always be added the defined way you should set the additional parameter FORCE YES to disable the use of `find_package`
+
```
TheLartians, line 204: Not sure this is easy to understand. How about instead adding the following at the end of the `CPM_USE_LOCAL_PACKAGES` section (after line 208)?

```markdown
Note that this does not apply to dependencies that have been defined with a truthy `FORCE` parameter. These will be added as defined.
```

TheLartians (commented):
Thanks for the update!

README.md
ClausKlein: done

TheLartians (approved):

Issue #460, opened 2023-03-07
Cannot pass `GIT_SUBMODULES ""` to disablesubmodule updates


CraigHutchinson:
This is due to using `UNPARSED_ARGUMENTS` is being used internally which removes empty values erroneously! 
https://gitlab.kitware.com/cmake/cmake/-/issues/24578

CraigHutchinson:
NOTE:  need to check if the 'internal' issue on FetchContent>ExternalProject changes can be applied to CPM as a fix: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4729/diffs

CraigHutchinson:
Most notably this change to use a potentially 'compatible' variant of `cmake_parse_arguments` using `PARSE_ARGV` TBC:
```
- cmake_parse_arguments(ARG "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+  cmake_parse_arguments(PARSE_ARGV 1 ARG
+   "${options}" "${oneValueArgs}" "${multiValueArgs}")
```

CraigHutchinson:
Validated the approach. It does work so will resolve the issue when `Cmake >= 3.7`
![image](https://user-images.githubusercontent.com/54269136/223430278-bae63dd3-adf4-4bd0-995b-cda9ca3130ef.png)

PR #461, opened 2023-03-07
Preserving forwarding of empty string arguments

CraigHutchinson:
# Background
This PR solves CPM issue https://github.com/cpm-cmake/CPM.cmake/issues/460

Empty-string arguments are not handled by Cmake well. There is a workaround here that is a hack for one specific parameter `GIT_SUBMODULES` https://gitlab.kitware.com/cmake/cmake/-/issues/20579 but there is also a proper fix to forwarding empty-string arguments applied under https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4729 which closed the Issue. CPM needs the same/similar also.

# This PR
This PR mirrors the changes in https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4729 so that CPM will forward all arguments onto `FetchContent`>`ExternalProject` unmodified when they contain empty string arguments

# Changes
- use `cmake_parse_arguments(PARSE_ARGV` which supports Empty-string arguments e.g. key-value pairs such as `GIT_SUBMODULES ""`
- Use `cmake_language(EVAL` for perfect-forwarding of empty-string argument lists

Also:
- Support cmake-format on Windows
- Auto fixes some whitespace

# Separate concern
For separate PR (future):
- `CPMFindPackage` not updated

# ToDO

- [x]  Fix for pre 3.18 usage of EVAL
- [x] Investigate/fix Test `package-lock_prettify` TYPO-Fixed
- [x] Add unit-test for empty string argument e.g. GIT_SUBMODULES ""


CraigHutchinson:
So the code using `EVAL` is CMake 3.18, this is currently failing. There is a legacy functionality in the docs but not sure it will handle empty strings. https://cmake.org/cmake/help/latest/command/cmake_language.html#evaluating-code

Potential resolution is to fix <3.18 by passing arguments the normal way and maybe warn when empty-values are provided as arguments as not-supported etc. TBC

CraigHutchinson:
CI Integration is failing due to:
```
/home/runner/work/CPM.cmake/CPM.cmake/build/integration/system_warnings/dependency_added_using_system-system-bin/_deps/adder-src/code/adder/adder.hpp:4:38:
    error: no return statement in function returning non-void [-Werror=return-type]
      4 | inline int emitWarning(int unused) { }
```

TheLartians:
> @TheLartians is this something CI related or some side-effect of this change?

From the diff of the PR it seems the `SYSTEM` option was removed (possibly a merge issue) which would cause CI to fail.

TheLartians (commented):
I'm having a hard time understanding this PR. To be sure, the issue is that it is currently impossible to disable recursive cloning using `FetchContent`?

- IMO the solution replacing the whole `CPMAddPackage` call with something using `eval` looks complicated and potentially risky for such a niche issue. Is there an easier way to handle it? E.g. in [here](https://gitlab.kitware.com/cmake/cmake/-/issues/20579) it was recommended to simply point the flag to a non-submodule directory instead. 
- Could you add a test case to check that it works as intended?

CraigHutchinson:
> * IMO the solution replacing the whole `CPMAddPackage` call with something using `eval` looks complicated and potentially risky for such a niche issue. Is there an easier way to handle it? E.g. in [here](https://gitlab.kitware.com/cmake/cmake/-/issues/20579) it was recommended to simply point the flag to a non-submodule directory instead.
20579 is unfortunately just a hack which doesn't fit with user expectations imho. CMake applied the fix to that in https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4729 which resolved this behavior for `FetchContent`>`ExternalProject` (among others) as the best UX option available. This PR mirrors the fix into CPM so that irrespective of there being new or other potential empty-string parameters they will all be correctly forwarded to the lower functions as the user expects. 

Internally using EVAL feels somewhat hacky but it appears to be the defacto workaround given the language limitations from what I can see!

> * Could you add a test case to check that it works as intended?
💯 I will be updating and provide additional tests soon.

cmake/CPM.cmake
```diff
@@ -505,17 +505,45 @@ function(cpm_override_fetchcontent contentName)
...
+  cpm_parse_add_package_single_arg("${arg}" __ARGN_multi)
 
-  list(LENGTH ARGN argnLength)
-  if(argnLength EQUAL 1)
-    cpm_parse_add_package_single_arg("${ARGN}" ARGN)
+  # The shorthand syntax implies EXCLUDE_FROM_ALL
+  list(APPEND __ARGN_multi EXCLUDE_FROM_ALL YES)
 
-    # The shorthand syntax implies EXCLUDE_FROM_ALL and SYSTEM
-    set(ARGN "${ARGN};EXCLUDE_FROM_ALL;YES;SYSTEM;YES;")
```
CraigHutchinson, line 517: Note: Change on this line was lost during merge - needs restore for `SYSTEM;YES`

CraigHutchinson: Done

cmake/CPM.cmake
```diff
@@ -531,15 +559,14 @@ function(CPMAddPackage)
       DOWNLOAD_COMMAND
       FIND_PACKAGE_ARGUMENTS
       NO_CACHE
-      SYSTEM
```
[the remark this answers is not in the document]
CraigHutchinson: Done

CraigHutchinson:
I currently have a 'minor' concern over the styling rules. Ideally I would say to prefer putting each Key-value pair on a separate line when their are multiple options. Justification is that this increases readability and especially change-diff is much more readable. I wonder if this is an option?...
image

CraigHutchinson:
I applied the same fix as is used under `package-lock_prettify.cmake` to disable cmake-format on the offending lines

CraigHutchinson:
Unit test added - Fails pre-changes except for `CPMDeclarePackage` which always worked. To check the internals specific functions are stubbed. This appears the only way to introspect `FetchContent_Declare` and others. It does feel like there may be other areas to ensure argument list is maintained that could be covered in future tests. The new tests focus here to cover the areas improved by this PR.

TheLartians:
Thanks for adding the test case! What do you think about using placeholder arguments instead as suggested [here](https://github.com/CraigHutchinson/CPM.cmake/pull/1)?

CraigHutchinson:
> Thanks for adding the test case! What do you think about using placeholder arguments instead of eval as suggested [here](https://github.com/CraigHutchinson/CPM.cmake/pull/1)?

That sounds great, I will take a deeper look really soon. It would be nice to not need Eval, albeit that is was used internally for CMake it doesn't mean it's the best and only option.

ClausKlein:
What is the state of this MR?

We have problems to prevent the cloning of git submodules using this [CMP0097](https://cmake.org/cmake/help/latest/policy/CMP0097.html#policy:CMP0097) CMake policy.

i.e.:
```cmake
# NOTE(CK): Set GIT_SUBMODULES to empty string to NOT initializes submodules!
cmake_policy(SET CMP0097 NEW)
CPMAddPackage(
  NAME project_options
  GIT_TAG v0.30.0
  GITHUB_REPOSITORY aminya/project_options
  # FIXME: GIT_SUBMODULES ""
  GIT_SUBMODULES "docs"
  # NOTE(CK): Workaround existing dir to prevent load of submodule
)
```

CraigHutchinson:
> 



> Thanks for adding the test case! What do you think about using placeholder arguments instead of eval as suggested [here](https://github.com/CraigHutchinson/CPM.cmake/pull/1)?

@TheLartians Appologies I have changed roles and not using CMake for a while ( 😢 ) and forgot to review the alternative approach.
May I propose that we use the approach wihtin this PR to unblock @ClausKlein and the non-eval alternative implementation could be reviewed as a separate PR improving this feature?

TheLartians:
I may be a bit too paranoid, but using a mechanism like eval for something simple like argument forwarding leaves me with a bad feeling. In my suggestion we essentially replace empty arguments using a placeholder string and remove it right before forwarding them.

I agree that it does add a bit more code and complexity than your version, but with proper docstrings and tests cases I hope that maintainability shouldn't be too big of an issue for this.

ClausKlein:
No!

ClausKlein:
What is about this MR?

Perhaps this helps to find a solution: [forwarding-command-arguments-in-cmake](https://crascit.com/2019/01/29/forwarding-command-arguments-in-cmake/#more-1005)

cmake/CPM.cmake
```diff
@@ -505,17 +505,50 @@ function(cpm_override_fetchcontent contentName)
...
+  else()
+    # Forward preserving empty string arguments
+    # (https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4729)
+    set(__ARGN_Quoted)
+    foreach(__ARG IN LISTS __ARGN)
+      string(APPEND __ARGN_Quoted " [==[${__ARG}]==]")
+    endforeach()
+    cpm_cmake_eval("cpm_add_package_multi_arg( ${__ARGN_Quoted} )")
+  endif()
+endmacro()
```
ClausKlein, line 533: There is, however, a specific case to be aware of which can result in unexpected behavior. Because macros treat their arguments as string substitutions rather than as variables, if they use ARGN in a place where a variable name is expected, the variable it will refer to will be in the scope from which the macro is called, not the ARGN from the macro’s own arguments. 

From chapter 9.2 [professional-cmake](https://crascit.com/professional-cmake/)

jdumas:
Hi. Have we reached a consensus regarding this PR yet? It seems we have either this version or [this alternative](https://github.com/CraigHutchinson/CPM.cmake/pull/1/) as a possibility? Personally if the upstream [FetchContent module](https://github.com/Kitware/CMake/blob/69a2eeb9dff5b60f2f1e5b425002a0fd45b7cadb/Modules/FetchContent.cmake#L1855) is using `cmake_language(EVAL ...)` I don't see why we couldn't do the same in CPM.

Without this feature it's currently impossible to disable cloning submodules with CPM, which can add quite a bit of extra things to download during project configuration.

TheLartians:
Yeah, I think going with the eval mechanism for now is ok, considering it is blocking an important feature. I still want to double-check the PR before, but unfortunately I do not have a lot of time at the moment.

jdumas:
Hi @TheLartians . I was wondering if there's any update on potentially merging this feature?

ClausKlein:
> Without this feature it's currently impossible to disable cloning submodules with CPM, which can add quite a bit of extra things to download during project configuration.

**There is a workaround** : set an `non existing` subdirectory to disable recursive cloning.

jdumas:
Oh nice, I missed this! I'll try that workaround, thanks!

TheLartians:
To revive the PR I've tried merging the new changes from main into it, however things broke and I haven't been able to resolve all issues yet. I'll try to get back to it soon.

Risto97:
I have also noticed this issue with GIT_SUBMODULES.
I am eagerly awaiting for the fix.
Thank you @TheLartians and @CraigHutchinson  for working on this!

jdumas:
Btw a little gotcha I ran into today: when using the `GIT_SUBMODULES "docs"` workaround proposed by @ClausKlein, make sure the `docs/` folder itself doesn't contain any submodule inside... Otherwise git will still attempt to clone them. I really hope we can get this PR or #687 integrated soon.

TheLartians:
As my changes have diverged quite a bit from the original PR I've moved them to #699, which also contains my initial idea using argument escaping. This probably wasn't all needed, but it helped me resolve the merge conflicts and get the test suite to pass. In the end I agree that `eval` seems to be the cleanest approach for forwarding the arguments to `FetchContent_Declare` in the end.

@CraigHutchinson could you take a look at my branch if it solves the use case sufficiently?

PR #463, opened 2023-03-20
Add AALTITOAD to "Built with CPM.cmake"

sillydan1:
I'm using CPM to manage (almost) all of my dependencies for my formal verification engine and associated projects. This enabled me to extract all of the individual libraries into separate repositories so they can be reused for other projects.

It's a small project, but I hope you guys don't mind having me on the list anyway.


TheLartians (approved):
Looks very interesting, thanks for sharing!

PR #468, opened 2023-04-15
if version=latest, use the last tagged version in repository

d-e-e-p:
eg:

```
CPMAddPackage(
  NAME fibonacci
  GIT_REPOSITORY https://github.com/cpm-cmake/testpack-fibonacci.git
  VERSION latest
)
```

motivation:

this is common case for our setup, where for modules like gh:TheLartians/PackageProject.cmake , we just want to pickup the latest version.  so i asked chatGPT:
prompt: how to download the latest tag of a package with CPMAddPackage
answer:

```
include(cmake/CPM.cmake)

CPMAddPackage(
  NAME <package_name>
  VERSION *
)
```

bing AI had something similar to say, so anyway I think this would be a useful enhancement.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -178,6 +178,26 @@ if(NOT CPM_DONT_CREATE_PACKAGE_LOCK)
   )
 endif()
 
+# look for string in TAG like "latest" or "master"
+function(cpm_fetch_git_tag_from_version_description URI RESULT)
+    find_package(Git REQUIRED)
+    if(NOT GIT_EXECUTABLE)
+        return()
+    endif()
+
+    # TODO: if not tag exists in rep, revert to last checkin
+    # TODO: make it work for case of multuple URI
+    # TODO: replace tail and sed with string functions so we don't have to call bash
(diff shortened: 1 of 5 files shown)


d-e-e-p:
version ranges would be very useful for any project following semantic versioning:
```
CPMAddPackage("gh:google/benchmark@1.7.*")
```
or 
```
CPMAddPackage("gh:google/benchmark@1.*")
```

anyway it might be useful to fetch available versions of a tool anyway for diagnostic messages on version mismatch.

by locking do you mean something like this?
https://docs.gradle.org/current/userguide/dependency_locking.html

TheLartians:
> by locking do you mean something like this?

Yes that or similar to the [package-lock.json](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json) used by npm. In fact we already have a [rudimentary package lock](https://github.com/cpm-cmake/CPM.cmake/blob/master/README.md#package-lock) setup, however there is currently no automation to keep it in sync with changing dependency versions in the CMakeLists, which would be needed for this feature.

PR #485, opened 2023-07-09
fix(lock): fix typo: CPM_CACHE_SOURCE -> CPM_SOURCE_CACHE

### LINKED ISSUE #484: Deadlock on file locking
With 0.38.0/1 i see deadlocks while CPM tries to lock a cache file. 0.37.0 is fine.

- [x] cpm cache used, but the folder must be deleted before (with cached folder it doe not happen)

```cmake
set(ENV{CPM_SOURCE_CACHE} ${CMAKE_CURRENT_LIST_DIR}/.cpm)
```
locations:
CPM_0.38.0.cmake(719)
CPM_0.38.1.cmake(721):

I guess its caused by: https://github.com/cpm-cmake/CPM.cmake/commit/0938e8fde7f5f9d2f086d4402aa82c7e47e37df5

vulptex:
fixes #484


TheLartians (approved):
Oops, thanks for fixing this!

PR #486, opened 2023-07-13
ci(cmake): split additional_commands to its own file

JohelEGP:
I'm an user of `cmake-format`.

I'd like to make use of the `additional_commands` defined in this repository's `.cmake-format`.
So I setup my build to copy it to `cmake/.cmake-format-additional_commands-cpm`.
Unfortunately, that includes the `format` portion.
And it seems that `cmake-format` processes those last, so my own `format` is overwritten.

In this PR, I split the `additional_commands` in `.cmake-format` to under `cmake/.cmake-format-additional_commands-cpm`.
To avoid polluting the top-level, I reuse existing `cmake` directory.
I prefer this approach to splitting because it makes updating the resulting config file a simple CMake `file(COPY_FILE)` call.

Now, users of `cmake-format` that want their calls to CPM functions well-formatted
just need to add `include: ["cmake/.cmake-format-additional_commands-cpm"]` to their `.cmake-format`.


TheLartians (approved):
Thanks for the PR! Splitting the code for CPM commands makes a lot of sense.

PR #504, opened 2023-09-04
chore(license): update license year

carlocorradini:
Update license year.

This process needs to be automated using a script that runs on January 1st and creates a new PR (any idea?)

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -5,7 +5,7 @@
 # MIT License
 # -----------
 #[[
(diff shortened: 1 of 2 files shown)


OlivierLDff:
Just remove the year and live with it
https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/

TheLartians (approved):

TheLartians:
Great thanks for noticing and updating this! For long term I agree with @OlivierLDff that it seems the year isn't really needed and would prefer simply removing to building a script for keeping it up to date.

Issue #505, opened 2023-09-13
FetchContent failure when using CPM on a single project with multiple subdirectories with cmake < 3.17


Anton3:
Related: #192

Minimal structure:

```
.
├── CMakeLists.txt
├── get_cpm.cmake
├── libA
│   └── CMakeLists.txt
└── libB
    └── CMakeLists.txt
```

Here, both `libA` and `libB` use the same `get_cpm.cmake` to download their third-party dependencies.

The symptoms are the same as in #192. The solution is the same as well, we just need to relax the conditions for the extra `include(FetchContent)`.

PR #506, opened 2023-09-13
fix(CPM.cmake): workaround FetchContent failure for cmake < 3.17

Anton3:
Fixes: #505

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -46,6 +46,13 @@ set(CURRENT_CPM_VERSION 1.0.0-development-version)
 
 get_filename_component(CPM_CURRENT_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}" REALPATH)
 if(CPM_DIRECTORY)
+  if(${CMAKE_VERSION} VERSION_LESS "3.17.0")
+    get_directory_property(hasParent PARENT_DIRECTORY)
+    if(NOT CPM_DIRECTORY STREQUAL CPM_CURRENT_DIRECTORY OR ${hasParent})
+      include(FetchContent)
+    endif()
+  endif()
+
   if(NOT CPM_DIRECTORY STREQUAL CPM_CURRENT_DIRECTORY)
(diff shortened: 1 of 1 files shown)


cmake/CPM.cmake
```diff
@@ -46,6 +46,13 @@ set(CURRENT_CPM_VERSION 1.0.0-development-version)
 
 get_filename_component(CPM_CURRENT_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}" REALPATH)
 if(CPM_DIRECTORY)
+  if(${CMAKE_VERSION} VERSION_LESS "3.17.0")
+    get_directory_property(hasParent PARENT_DIRECTORY)
+    if(NOT CPM_DIRECTORY STREQUAL CPM_CURRENT_DIRECTORY OR ${hasParent})
```
TheLartians, line 51: Could you add a comment above this explaining what issue this check fixes? 

Also I feel some parenthesis could be useful around the `NOT` expression as I'm confused by CMake's operator precedence.

TheLartians, line 51: Additionally the test cases seem to fail if `hasParent` is an empty string.

```
1: CMake Error at /home/runner/work/CPM.cmake/CPM.cmake/cmake/CPM.cmake:51 (if):
1:   if given arguments:
1: 
1:     "NOT" "CPM_DIRECTORY" "STREQUAL" "CPM_CURRENT_DIRECTORY" "OR"
1: 
1:   Unknown arguments specified
1: Call Stack (most recent call first):
```

It should be fixable by adding quotation marks.

```suggestion
    if(NOT CPM_DIRECTORY STREQUAL CPM_CURRENT_DIRECTORY OR "${hasParent}")
```

TheLartians (changes requested):
Hey thanks for the issue and accompanying fix! Would it make sense to also add a [test case](https://github.com/cpm-cmake/CPM.cmake/blob/master/test/integration/README.md) for this, so we don't accidentally break it in the future?

PR #507, opened 2023-09-18
Fix new release script to use correct version and hash sum

TheLartians:
This fixes an issue with the new download script from #474, discovered after testing, where the hash and version were not properly replaced by the release script workflow. Additionally, it now only logs on errors instead of on each successful run of the download command.


TheLartians:
Will merge without proper review, as the issues are blocking the next release.

PR #511, opened 2023-10-04
Fix: Include Missing Arguments in `cpm_prettify_package_arguments` Function

### LINKED ISSUE #510: Incorrect Package Lock Formatting After Second Update
I'm using the latest version of CPM.cmake (v0.38.5).

When I add a package with `OPTIONS` specified to my project and run the `cpm-update-package-lock` target, it initially generates the package lock like this:

```cmake
# CPM Package Lock
# This file should be committed to version control

# yaml-cpp
CPMDeclarePackage(yaml-cpp
  GIT_TAG 0.8.0
  GITHUB_REPOSITORY jbeder/yaml-cpp
  SYSTEM YES
  OPTIONS
    "YAML_CPP_FORMAT_SOURCE OFF"
  EXCLUDE_FROM_ALL YES
)
```

However, upon running the `cpm-update-package-lock` target for the second time, it updates the package lock incorrectly, resulting in this formatting:

```cmake
# CPM Package Lock
# This file should be committed to version control

# yaml-cpp
CPMDeclarePackage(yaml-cpp
  GIT_TAG 0.8.0
  GITHUB_REPOSITORY jbeder/yaml-cpp
  SYSTEM YES
  OPTIONS
    "YAML_CPP_FORMAT_SOURCE OFF"
    "EXCLUDE_FROM_ALL"
    "YES"
)
```

You can see that the `EXCLUDE_FROM_ALL` option is now mistakenly combined with the other options. This behavior seems to be due to the interpretation of `EXCLUDE_FROM_ALL` as part of the `OPTIONS`.

threeal:
This pull request added missing arguments from `CPMAddPackage` function which is not listed in the `cpm_prettify_package_arguments`. This modification aims to resolve problems related to incorrect package lock formatting. It resolves #510.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -1095,12 +1095,15 @@ function(cpm_prettify_package_arguments OUT_VAR IS_IN_COMMENT)
       DOWNLOAD_ONLY
       GITHUB_REPOSITORY
       GITLAB_REPOSITORY
+      BITBUCKET_REPOSITORY
       GIT_REPOSITORY
       SOURCE_DIR
       FIND_PACKAGE_ARGUMENTS
       NO_CACHE
       SYSTEM
       GIT_SHALLOW
+      EXCLUDE_FROM_ALL
+      SOURCE_SUBDIR
   )
   set(multiValueArgs URL OPTIONS DOWNLOAD_COMMAND)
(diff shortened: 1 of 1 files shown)


TheLartians (approved):
Great, thanks for identifying the issue and providing a fix!

PR #522, opened 2023-11-20
Treat relative paths as relative to parent project's remote

### LINKED ISSUE #520: Honour new relative git URL behaviour controlled by CMP0150
I added support for relative git URLs to FetchContent/ExternalProject which was introduced in cmake v3.27 under policy [CMP0150](https://cmake.org/cmake/help/latest/policy/CMP0150.html). Unfortunately, this doesn't seem to work when wrapped with CPM and it looks like this is because CPM minimum cmake version is set to 3.14.

I'm not sure on the best strategy here, but one option would be for CPM to inherit the policies from the parent project, and just assert that they're set at least to a minimum of 3.14 as CPM needs, rather than actually explicitly setting them to 3.14 as is currently the case. Alternatively, it should at least honour policy 150 and pass it through if it can.

Happy to submit a PR for this if there's a preferred solution.

inkychris:
Closes #520.


TheLartians (approved):
Hey, thanks for the pull request and even adding a test case! 

I've refactored the test so it uses relative URLs defined in a different project ([this branch of testpack-fibadder](https://github.com/cpm-cmake/testpack-fibadder/blob/v1.1.0-relative-urls/CMakeLists.txt)), so it won't break the tests on third-party forks of CPM.cmake.

After the change the test cases failed on CI, as it turns out the feature was introduced in CMake 3.27 but we were testing on 3.25 and 3.16. Not sure why they were previously passing, but we've updated our test runners and now seems to work fine.

inkychris:
Thanks for getting this merged! I realised it probably should have been an integration test instead of a unit-test but picked the easiest thing to copy/paste 😅

PR #527, opened 2023-12-12
Feature/parsing arguments robustly

ClausKlein:
Parsing Arguments Robustly
    
    Prevent test errors if CPM_SOURCE_CACHE is set
    
    Use cmake_parse_arguments(PARSE_ARGV...)
    
    Add CMake workflow presets test


ClausKlein:
I give it up, the CPM.cmake interface has no long term successs!

PR #532, opened 2024-01-14
Re-export `CMAKE_MODULE_PATH`

threeal:
This pull request re-export the `CMAKE_MODULE_PATH` variable to the parent scope from the `add_subdirectory` function. This function is useful because some packages, like Catch2, export the `CMAKE_MODULE_PATH` of their script files to the parent scope, as could be seen here:

[image](https://github.com/catchorg/Catch2/blob/863c662c0eff026300f4d729a7054e90d6d12cdd/CMakeLists.txt#L45-L53)

While this implementation may be somewhat unconventional and a bit hacky, I don't know of any better ways to solve this issue. It allows script files to be directly used instead of having to delve deep into the project source files. Previously, if we wanted to use the `catch_discover_tests` function with Catch2, we had to do the following:

```cmake
CPMAddPackage("gh:catchorg/Catch2@3.4.0")
include("${Catch2_SOURCE_DIR}/extras/Catch.cmake")
```

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -756,6 +756,7 @@ function(CPMAddPackage)
         "${CPM_ARGS_SYSTEM}"
         "${CPM_ARGS_OPTIONS}"
       )
+      set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} PARENT_SCOPE)
       set(PACKAGE_INFO "${PACKAGE_INFO} at ${download_directory}")
 
       # As the source dir is already cached/populated, we override the call to FetchContent.
@@ -815,6 +816,7 @@ function(CPMAddPackage)
         "${CPM_ARGS_SYSTEM}"
         "${CPM_ARGS_OPTIONS}"
       )
+      set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} PARENT_SCOPE)
     endif()
     cpm_get_fetch_properties("${CPM_ARGS_NAME}")
   endif()
@@ -986,6 +988,7 @@ function(
     set(CPM_OLD_INDENT "${CPM_INDENT}")
     set(CPM_INDENT "${CPM_INDENT} ${PACKAGE}:")
     add_subdirectory(${SOURCE_DIR} ${BINARY_DIR} ${addSubdirectoryExtraArgs})
+    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} PARENT_SCOPE)
     set(CPM_INDENT "${CPM_OLD_INDENT}")
   endif()
 endfunction()
--- a/examples/catch2/CMakeLists.txt
+++ b/examples/catch2/CMakeLists.txt
@@ -17,5 +17,7 @@ target_compile_features(CPMExampleCatch2 PRIVATE cxx_std_17)
 
 # ---- Enable testing ----
 
+include(Catch)
+
 enable_testing()
-add_test(CPMExampleCatch2 CPMExampleCatch2)
+catch_discover_tests(CPMExampleCatch2)


TheLartians (commented):
Hey thanks for the PR, I wasn't aware of this mechanism in packages! I don't see how this can break anything, so in general the implementation seems fine. Some ideas for improvement:

- Can you add a comment in `cpm_add_subdirectory` explaining why the modules are exported to the parent scope there?
- Would it be possible to move the parent scope export of the module path to the `cpm_export_variables` macro? That way we don't need to remember to call it after each `cpm_add_subdirectory`
- To avoid regressions, do you think there's an easy way we could add an integration test for this?

aedart:
Hi @TheLartians and @threeal 

I too have some issues with my [package](https://github.com/rsps/cmake-scripts), in regards to appending a module path to `CMAKE_MODULE_PATH`.
Basically, whenever I install the package using cmake's native `FetchContent_Declare`, it appears to work (_which might just be luck - only works if I use FetchContent from my consuming project's root level `CMakeLists.txt` file._). This is most likely because `FetchContent_MakeAvailable()` appears to include my package's `CMakeLists.txt` via `add_subdirectory()`.
```cmake
include(FetchContent)
FetchContent_Declare(
    "rsp-cmake-scripts"
    GIT_REPOSITORY "https://github.com/rsps/cmake-scripts"
    GIT_TAG "43febe92df4a06a31338e380eb2dc64683f8a273" # Commit hash for now, version not yet released.
)
FetchContent_MakeAvailable("rsp-cmake-scripts")

include("rsp/debug") # Works
dump(CMAKE_MODULE_PATH) # Works
```

But, when trying to fetch my package using CPM, any changes to `CMAKE_MODULE_PATH` are not propagated out of the `CPMAddPackage()` function. E.g.:

```cmake
CPMAddPackage(
    NAME "rsp-cmake-scripts"
    GIT_TAG "43febe92df4a06a31338e380eb2dc64683f8a273" # Commit hash for now, version not yet released.
    # VERSION "${RSP_CMAKE_SCRIPTS_VERSION}"
    GITHUB_REPOSITORY "rsps/cmake-scripts"
    OPTIONS "RSP_ENABLE_ANSI true"
)

include("rsp/debug") # CMake Error: include could not find requested file: rsp/debug
dump(CMAKE_MODULE_PATH) # ...not reached
```

## Experiment A (failed)

I have locally attempted to set the `CMAKE_MODULE_PATH` in the parent scope, by modifying the `cpm_export_variables()` macro:
```cmake
macro(cpm_export_variables name)
  # ... previous not shown ...

  # Attempt to propagate eventual appended CMAKE_MODULE_PATH paths to parent
  set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}" PARENT_SCOPE)
endmacro()
```

Sadly, the above change had no effect. My package's module path isn't available.

## Experiment B

Same as "Experiment A", but with a modification to `cpm_fetch_package()`:

```cmake
function(cpm_fetch_package PACKAGE DOWNLOAD_ONLY populated)
  # ... previous not shown ...

  set(${PACKAGE}_SOURCE_DIR
      ${${lower_case_name}_SOURCE_DIR}
      PARENT_SCOPE
  )
  set(${PACKAGE}_BINARY_DIR
      ${${lower_case_name}_BINARY_DIR}
      PARENT_SCOPE
  )

  # Attempt to propagate eventual appended CMAKE_MODULE_PATH paths to parent - allows
  # cpm_export_variables() to propagate change upwards...
  set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}" PARENT_SCOPE)
endfunction()
```

This seems to work in my case, yet I'm not sure if other CPM functions need to be modified?

For now, I do not see any viable work-around on how to make this work as desired. It is not uncommon for packages, libraries, ...etc, to append paths to `CMAKE_MODULE_PATH`, such that consuming projects are able to make use of cmake utilities, custom paths, or resources that are somehow made available.

In any case, I really hope that CPM can make this work.

aedart:
In addition to the previous comment, the problem is actually not just for appended paths to `CMAKE_MODULE_PATH`, but basically every kind of variable that a "package" attempts to expose in a parent scope!

I'm not sure what a correct solution would be (_I'm not a CMake expert... I'm still struggling to understand the basics_).

PR #546, opened 2024-03-11
feature: allow additional arguments after shorthand syntax

SGSSGene:
This allows to combine the shorthand syntax with additional arguments:
```
CPMAddPackage("gh:nlohmann/json@3.9.1" OPTIONS "JSON_BUildTests OFF")
```

This is much shorter than the longer syntax way of writing:
```
CPMAddPackage(
  NAME nlohmann_json
  VERSION 3.9.1
  OPTIONS
    "JSON_BuildTests OFF"
)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ If an additional optional parameter `SYSTEM` is set to a truthy value, the SYSTE
 See the [add_subdirectory ](https://cmake.org/cmake/help/latest/command/add_subdirectory.html?highlight=add_subdirectory)
 and [SYSTEM](https://cmake.org/cmake/help/latest/prop_tgt/SYSTEM.html#prop_tgt:SYSTEM) target property for details.
 
-A single-argument compact syntax is also supported:
+A shorthand syntax is also supported:
 
 ```cmake
 # A git package from a given uri with a version
@@ -109,6 +109,13 @@ CPMAddPackage("https://example.com/my-package-1.2.3.zip")
 CPMAddPackage("https://example.com/my-package-1.2.3.zip#MD5=68e20f674a48be38d60e129f600faf7d")
 # An archive package from a given url. The version is explicitly given
 CPMAddPackage("https://example.com/my-package.zip@1.2.3")
+```
+
+Additionally, the shorthand syntax can be combined with the other options from above:
+```cmake
+CPMAddPackage("gh:nlohmann/json@3.9.1"
+  OPTIONS "JSON_BuildTests OFF"
+)
 ```
 
 After calling `CPMAddPackage`, the following variables are defined in the local scope, where `<dependency>` is the name of the dependency.
--- a/README.md
+++ b/README.md
@@ -403,12 +403,8 @@ CPMAddPackage("gh:jbeder/yaml-cpp#yaml-cpp-0.6.3@0.6.3")
 ### [nlohmann/json](https://github.com/nlohmann/json)
 
 ```cmake
-CPMAddPackage(
-  NAME nlohmann_json
-  VERSION 3.9.1
-  GITHUB_REPOSITORY nlohmann/json
-  OPTIONS
-    "JSON_BuildTests OFF"
+CPMAddPackage("gh:nlohmann/json@3.9.1"
+  OPTIONS "JSON_BuildTests OFF"
 )
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -397,12 +397,7 @@ CPMAddPackage("gh:jbeder/yaml-cpp#yaml-cpp-0.6.3@0.6.3")
...
 
 ```cmake
-CPMAddPackage(
-  NAME nlohmann_json
-  VERSION 3.9.1
-  GITHUB_REPOSITORY nlohmann/json
-  OPTIONS
-    "JSON_BuildTests OFF"
+CPMAddPackage("gh:nlohmann/json@3.9.1" OPTIONS "JSON_BuildTests OFF")
 )
```
TheLartians, line 401: I actually find this slightly harder to read than before, maybe it would be better to keep some of the indentation?

```suggestion
CPMAddPackage("gh:nlohmann/json@3.9.1"
  OPTIONS "JSON_BuildTests OFF"
)
```

TheLartians (changes requested):
Hey, thanks for the PR! In general I like the idea of mixing the short syntax with additional arguments and think this is a good approach in doing so. Could you perhaps add an integration test case for this?

Also I'm not sure but it could be that the cmake-format config might need to be updated for this new syntax.

SGSSGene:
> Hey, thanks for the PR! In general I like the idea of mixing the short syntax with additional arguments and think this is a good approach in doing so. Could you perhaps add an integration test case for this?
> 
> Also I'm not sure but it could be that the cmake-format config might need to be updated for this new syntax.

Yes, there are a few things that need to be updated. I wanted to discuss, if there is any chance on success before investing to much time.
I will look through everything and update this PR

test/integration/test_simple.rb
```diff
@@ -83,9 +83,38 @@ def test_update_single_package
...
+      # * using-adder - for this project
+      # ...and notably no test for adder, which must be disabled from the option override from above
+      assert_equal ['simple', 'using-adder'], exes
+
+    }
 
     create_with_commit_sha.()
     update_to_version_1.()
     update_with_option_off_and_build.()
+  update_with_option_off_and_build_with_shorthand_syntax.()
```
TheLartians, line 118: I feel like having this run as part of a larger test makes reasoning about the test itself difficult. Do you think you could move checking the shorthand extra arguments into its own isolated test case?

TheLartians (commented):
Thanks, the changes look good, just want to make sure that the test actually does what it should. One thing that I notice from the readme changes is that having a mix of single + named arguments looks quite inconsistent and therefore harder to read.

How would you feel if we instead make a compromise by introducing a `URI` argument that takes the role of the single argument instead?

This would allow us to do

```cmake
CPMAddPackage(
  URI "gh:nlohmann/json@3.9.1"
  OPTIONS "JSON_BuildTests OFF"
)
```

Instead of

```cmake
CPMAddPackage("gh:nlohmann/json@3.9.1"
  OPTIONS "JSON_BuildTests OFF"
)
```

To me this looks much nicer, even though I know CMake itself likes to use this mix of syntax. Of course, using only the single argument syntax should continue to be valid.

SGSSGene:
> Thanks, the changes look good, just want to make sure that the test actually does what it should. One thing that I notice from the readme changes is that having a mix of single + named arguments looks quite inconsistent and therefore harder to read.
> 
> How would you feel if we instead make a compromise by introducing a `URI` argument that takes the role of the single argument instead?
> 
> This would allow us to do
> 
> ```cmake
> CPMAddPackage(
>   URI "gh:nlohmann/json@3.9.1"
>   OPTIONS "JSON_BuildTests OFF"
> )
> ```
> 
> Instead of
> 
> ```cmake
> CPMAddPackage("gh:nlohmann/json@3.9.1"
>   OPTIONS "JSON_BuildTests OFF"
> )
> ```
> 
> To me this looks much nicer, even though I know CMake itself likes to use this mix of syntax. Of course, using only the single argument syntax should continue to be valid.

I am not so sure of the `URI` idea.
Pro:
 - It does look nice.
  
Con:
- Not easy extendible from shorthand syntax, my main idea was that I can just slap on further 'OPTIONS', with this I also need to add a URI
- Harder to document. Currently we have two cases: 1. normal usage, 2. shorthand syntax. With `URI` we kind of have to split the 'normal usage' into two cases. I don't see how this can be documented in any understandable way. (Assuming we don't delete the 2. shorthand syntax)
- Also I think it is harder to implement (not so sure about this)

I like to think about it similar to how `target_link_libraries` works. First I set a free given parameter, and than I add all the other with `PRIVATE`,`PUBLIC` or `INTERFACE`.

test/integration/test_simple.rb
SGSSGene: Sure :+1:

Richardk2n:
Gentleman, how is this coming along?
Is there anything preventing this from getting merged other than the `URI` or not `URI` question?

SGSSGene:
The question of `URI` or not `URI` is up to @TheLartians . Both is fine to me, @TheLartians needs to decide what he would like.

The only big stopper for the `URI` idea for me is, I think `URI` is a bad keyword, since it is not a URI?!
Any argument for saying it is a URI or an alternative keyword would be helpful.

TheLartians:
Thanks for the alternative implementation! I personally prefer that version a lot and added some comments in the other PR.

SGSSGene:
droped this PR in favor of #617

PR #549, opened 2024-03-16
Add xxHash example

### LINKED ISSUE #416: Example for xxhash
I am trying to bring in xxhash with CPM (0.32) but I can't figure it out.

xxhash has an un-official cmake support, with a subfolder, that I got to work in the past with the SOURCE_SUBDIR magic value. Is *SOURCE_SUBDIR* supported with CPM ?

I am also trying to do it the DOWNLOAD_ONLY way, and then create a header only library.

```
if(xxhash_ADDED)
  add_library(xxhash INTERFACE)
  target_sources(xxhash INTERFACE ${xxhash_SOURCE_DIR}/xxhash.h ${xxhash_SOURCE_DIR}/xxh3.h)
  
  packageproject(...)
endif()

but I get lots of errors later on.

```
CMake Error in CMakeLists.txt:
  Target "xxhash" INTERFACE_SOURCES property contains path:

    "/Users/benjamin.sergeant/projects/xyz/build/_deps/xxhash-src/xxh3.h"

  which is prefixed in the build directory.
```

### PR DESCRIPTION
**Pull request by RozeFound**
closes #416

### DIFF
--- a/examples/xxHash/CMakeLists.txt
+++ b/examples/xxHash/CMakeLists.txt
@@ -7,12 +7,10 @@ project(CPMxxHashExample)
 include(../../cmake/CPM.cmake)
 
 CPMAddPackage(
-    GITHUB_REPOSITORY Cyan4973/xxHash
-    GIT_TAG v0.8.2
-    OPTIONS 
-        "XXHASH_BUILD_ENABLE_INLINE_API OFF"
-        "XXHASH_BUILD_XXHSUM OFF"
-    SOURCE_SUBDIR cmake_unofficial
+  GITHUB_REPOSITORY Cyan4973/xxHash
+  GIT_TAG v0.8.2
+  OPTIONS "XXHASH_BUILD_ENABLE_INLINE_API OFF" "XXHASH_BUILD_XXHSUM OFF"
+  SOURCE_SUBDIR cmake_unofficial
 )
 
 # ---- Executable ----
--- a/examples/xxHash/CMakeLists.txt
+++ b/examples/xxHash/CMakeLists.txt
@@ -1,0 +1,22 @@
+cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
+
+project(CPMxxHashExample)
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by TheLartians — COMMENTED** [review]
Hey, thanks for adding the example, this looks like a useful library! 
It seems however that it is failing to build on our CI. Could you double-check the code? The error is the following:

```
  CMake Deprecation Warning at /home/runner/work/CPM.cmake/CPM.cmake/build/xxHash/_deps/xxhash-src/cmake_unofficial/CMakeLists.txt:8 (cmake_minimum_required):
    Compatibility with CMake < 3.5 will be removed from a future version of
    CMake.
  
    Update the VERSION argument <min> value or use a ...<max> suffix to tell
    CMake that the project does not need compatibility with older versions.
  
  
  CMake Error at CMakeLists.txt:22 (target_link_libraries):
    Target "CPMxxHashExample" links to:
  
      xxHash::xxHash
  
    but the target was not found.  Possible reasons include:
  
      * There is a typo in the target name.
      * A find_package call is missing for an IMPORTED target.
      * An ALIAS target is missing.
  
  
  
  CMake Generate step failed.  Build files cannot be regenerated correctly.
```

Also be sure to run cmake-format. For more info, check out our [style guides](https://github.com/cpm-cmake/CPM.cmake/blob/master/CONTRIBUTING.md#styleguides).

RozeFound:
I (hopefully) fixed link error, but unfortunately I can't get cmake-format to work for some reason.
Do I also need to fix the warning? It's fixed in one of a post release commits, so we can simply use it instead of a release tag.

TheLartians:
> I fixed link error, but unfortunately I can't get cmake-format to work for some reason.

Ok, no worries, I've run the code formatters on my machine.

> Do I also need to fix the warning? It's fixed in one of a post-release commits, so we can simply use this instead of a release tag.

I like using release tags here, so for now let's live with the warning and update once there is another release.

TheLartians (approved):
Thanks for adding the example!

PR #555, opened 2024-04-09
Remove 'bug' from comment

PercentBoat4164:
This is a super simple change. The reason for it is that CLion, and probably other IDEs, will report this comment as a bug when looking at the TODOs of a project that indexes the `CPM.cmake` file. Avoiding the word `bug` prevents that.

I know that this is a stupid pull request, but it was ***bug***ging me.


PercentBoat4164:
Oops. I messed up the style twice. I think it was the line endings that got me. That is what I get for thinking I could do it in GitHub's editor.

TheLartians (approved):
Haha I've never heard about such behaviour, but am completely fine with the rephrasing! Thanks!

Issue #523, opened 2023-11-27
wxWidgets example


idotta:
Hi there! Would someone please show me an example on how to use CPM to build a project using wxWidgets?

I couldn't make it work and I'm not experienced with CMake. I know there's something missing, but I can't figure out what. I'm trying to do it simply by using:

```
CPMAddPackage("gh:wxWidgets/wxWidgets@3.2.4")
include(${wxWidgets_USE_FILE})
add_executable(app main.cpp)
target_link_libraries(app PRIVATE ${wxWidgets_LIBRARIES})
```

jolz:
Hi. you missed to give some error message...

The following snippet works for me (linux, all required -dev-Packages installed).

```cmake
cmake_minimum_required(VERSION 3.24)

include(${CMAKE_SOURCE_DIR}/cmake/CPM.cmake)
CPMAddPackage("gh:wxWidgets/wxWidgets@3.2.5")

project(hello CXX)
add_executable(${PROJECT_NAME} WIN32 hello.cpp)
target_link_libraries(${PROJECT_NAME} PRIVATE wx::core)
```

idotta:
Hi there! Thank you so much for taking the time for a complete example!

PR #561, opened 2024-05-31
add wxWidgets example

jolz:
wxWidgets example, see  #523.

examples are based on wxWidgets official examples.

### DIFF
--- a/examples/wxWidgets/CMakeLists.txt
+++ b/examples/wxWidgets/CMakeLists.txt
@@ -1,0 +1,21 @@
+cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
+
+project(wxWidgetsExample)
+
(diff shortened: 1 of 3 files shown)


TheLartians (commented):
Thanks for adding the wxWidgets example! Unfortunately it doesn't compile on our CI, as we don't include the required `GTK3` framework. As we already have a plenty examples and this one requires additional setup steps, how about adding it to the entry to our [snippets wiki](https://github.com/cpm-cmake/CPM.cmake/wiki/More-Snippets) instead?

jolz:
[ok, done](https://github.com/cpm-cmake/CPM.cmake/wiki/More-Snippets#wxwidgets).
We could include the console-only example (pull request updated), but I am unsure if this makes sense since "wx without gui" feels a bit niche.

PR #578, opened 2024-07-27
Switch to non-deprecated FetchContent_Populate syntax (research PR)

TheLartians:
My own attempt at figuring out why #570 doesn't seem to satisfy our test cases and trying to understand what the differences between old and new syntax are. For one it seems that `FetchContent_GetProperties` does not seem to give us the SOURCE and BINARY directory, as I would have expected, so we need to implement the parsing ourselves. Now it seems that the handling of git has changed, leading to issues with the relative URL and deleted source dir test cases.


TheLartians:
Closing as the new syntax seems to come with more side-effects than expected.

PR #579, opened 2024-07-30
Add working sqlite example.

ScottBailey:
This is a very simple example that adds sqlite from an amalgamation source.


examples/sqlite/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+include(../../cmake/CPM.cmake)
+
+# sqlite - You know it, you love it!
+CPMAddPackage(
+  NAME sqlite
+  URL https://www.sqlite.org/2024/sqlite-amalgamation-3460000.zip
+  URL_HASH SHA256=712a7d09d2a22652fb06a49af516e051979a3984adb067da86760e60ed51a7f5
+  PATCHES sqlite.patch
+)
+
```
TheLartians, line 16: Thanks a lot for the extra example, this is an interesting use of the patches command! I'm a bit torn if we should encourage the use of the patches like this as the command still [seems to have](https://github.com/cpm-cmake/CPM.cmake/issues/577) [some issues](https://github.com/cpm-cmake/CPM.cmake/issues/618). Until the command is more mature I think we should still encourage the old way of adding the library without using patches:

```suggestion
if (sqlite_ADDDED)
  add_library(
    sqlite STATIC 
      ${sqlite_SOURCE_DIR}/sqlite3.c
      ${sqlite_SOURCE_DIR}/sqlite3.h 
      ${sqlite_SOURCE_DIR}/sqlite3ext.h
  )
  target_include_directories(sqlite PUBLIC SYSTEM "${sqlite_SOURCE_DIR}")
endif()
```

Issue #576, opened 2024-07-27
Incorrect find patch program under win32


crackedmind:
```cmake
    if(GIT_EXECUTABLE) # C:/Program Files/Git/cmd/git.exe
      get_filename_component(extra_search_path ${GIT_EXECUTABLE} DIRECTORY)
      get_filename_component(extra_search_path ${extra_search_path} DIRECTORY)
      get_filename_component(extra_search_path ${extra_search_path} DIRECTORY)
      # and here we got C:/Program Files instead of C:/Program Files/Git
      find_program(PATCH_EXECUTABLE patch HINTS "${extra_search_path}/usr/bin")
    endif()
```

ScottBailey:
With a little more info, this can be an easy fix! 

How did you install git? I mean from where?
And where is the patch command on your system?

Thanks!

crackedmind:
git version 2.45.2.windows.1
installed from official installer to C:\Program Files\Git
patch.exe location of  is C:\Program Files\Git\usr\bin
git.exe location is C:\Program Files\Git\cmd\git.exe

PR #581, opened 2024-07-31
Additional search path for patch.exe

ScottBailey:
It seems the windows installation locations of `patch.exe` and `git.exe` have either changed or are not fixed relative to one another (possibly based on install options?). Regardless, this change will search both the parent AND the parent-parent path for `patch.exe`.

Written to address #576.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -477,13 +477,13 @@ function(cpm_add_patches)
   find_program(PATCH_EXECUTABLE patch)
   if(WIN32 AND NOT PATCH_EXECUTABLE)
     # The Windows git executable is distributed with patch.exe. Find the path to the executable, if
-    # it exists, then search `../../usr/bin` for patch.exe.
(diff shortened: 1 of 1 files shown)


ScottBailey:
@crackedmind Please review and test this change on your system. I am unable to test it on mine.

crackedmind:
@ScottBailey it's working.

ScottBailey:
@TheLartians It looks like finding `patch.exe` is more dynamic than I initially thought. It changed! This PR makes it a bit better.

I had considered using a recursive file glob, but this is a much simpler change that will succeed or fail much faster. If the directory structure continues to change, we can revisit file glob.

Closes #576

Thank you for reporting @crackedmind

TheLartians (approved):
Thanks for the PR @ScottBailey, looks like a pragmatic fix! 
Also thanks @crackedmind for reporting and testing this.

TheLartians:
Fix is deployed in `v0.40.2` 🎉

PR #583, opened 2024-08-03
Update Boost example to v1.85.0

ScottBailey:
Basic update to 1.85.0. Most new or less advanced users probably want to start with this version. CMake support within Boost is a bit more mature.


ScottBailey:
1.86 is available. Closing.

PR #592, opened 2024-08-29
Update boost to 1.86

ScottBailey:
Basic update to 1.86.0. Most new or less advanced users probably want to start with version 1.85 or greater due to a greater maturity in Boost's CMake files.


TheLartians (approved):
Great, thanks for the update!

PR #593, opened 2024-08-30
Fetch content

flagarde:
Hi,

This is a rework for #446. This PR is taking into account 3 scenari :

1) User use the [get_cpm.cmake](https://github.com/cpm-cmake/CPM.cmake/blob/master/cmake/get_cpm.cmake) (no change for them)

2) Fetching with URL https://github.com/cpm-cmake/CPM.cmake/archive/refs/tags/v0.40.2.zip (use the .git_archival.txt)

3) Fetching with GIT_REPOSITORY (use git program to parse version)

As this PR doesn't modify the standard way to obtain CPM it allows to test it before switching to this method as the prefered one


TheLartians (dismissed):
Thanks for creating this PR! I really like this approach for "trying out" the this new way of obtaining CPM. Once tested with a new release we can add this approach as an alternative to the readme. I think it still needs a run cmake-format run before merging though.

flagarde:
@TheLartians I run cmake-format. Seems it is passing now

TheLartians (approved):
Great, thanks!

TheLartians:
Made a release `v0.40.7` for testing!

flagarde:
It seems to work but :
* Sha5 or sha256 should appear in the release page to be able to use it for FetchContent

* If the user use GIT_TAG even a release one v0.4.7 for example, it is still considered as development because .git present. One solution (if this is not the expected behaviour) is to parse the version extraced from git and if it corespond to a semver then not add the -development string

TheLartians:
> Sha5 or sha256 should appear in the release page to be able to use it for FetchContent

I think we could update the publish script to also create a `get_cpm_fetchcontent.cmake` to help with that at least.

> If the user use GIT_TAG even a release one v0.4.7 for example, it is still considered as development because .git present. One solution (if this is not the expected behaviour) is to parse the version extraced from git and if it corespond to a semver then not add the -development string

Yeah as the output of `git describe --tags` should also give the number of commits since the latest tag (if there are any), IMO it should be fine to omit the -development string as there should be no confusion with release versions.

flagarde:
For the sha256 I would do the contrary. Use some github action to write the SHA256 etc (all the one CMake can understand) inside the readme of the release. Modify the get_cpm to internally use this new method (it is more robust than file(download) cmake command).

PR #604, opened 2024-09-24
Add find_package overrides that also work in CONFIG mode (#498)

### LINKED ISSUE #498: OVERRIDE_FIND_PACKAGE doesn't work
I think because CPM avoids calling FetchContent_MakeAvailable(?), the OVERRIDE_FIND_PACKAGE argument to FetchContent_Declare doesn't work. The redirect files don't get created.

patstew:
CPM currently adds a FindFoo.cmake to the module path to override subsequent find_package calls. This makes sure that works when the find_package call is in CONFIG mode as well. Some of my dependencies use this in their CMakeLists.txt. This is equivalent to the OVERRIDE_FIND_PACKAGE argument to FetchContent, we could add that option to CPMAddPackage, but it seems like this is the intended default behaviour of CPM?

The include(....-extra.cmake) bits are something FetchContent does that's also useful here. They should maybe be added to the Find*.cmake version too?

Fixes #498

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -162,7 +162,7 @@ set(CPM_SOURCE_CACHE
     CACHE PATH "Directory to download CPM dependencies"
 )
 
-if(NOT CPM_DONT_UPDATE_MODULE_PATH)
+if(NOT CPM_DONT_UPDATE_MODULE_PATH AND NOT DEFINED CMAKE_FIND_PACKAGE_REDIRECTS_DIR)
   set(CPM_MODULE_PATH
       "${CMAKE_BINARY_DIR}/CPM_modules"
       CACHE INTERNAL ""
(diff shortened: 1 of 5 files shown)


cmake/CPM.cmake
```diff
@@ -269,10 +269,21 @@ endfunction()
 # finding the system library
 function(cpm_create_module_file Name)
   if(NOT CPM_DONT_UPDATE_MODULE_PATH)
-    # erase any previous modules
-    file(WRITE ${CPM_MODULE_PATH}/Find${Name}.cmake
-         "include(\"${CPM_FILE}\")\n${ARGN}\nset(${Name}_FOUND TRUE)"
-    )
+    if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24.0")
+      string(TOLOWER ${Name} NameLower)
```
TheLartians, line 273: Would it make sense to add a comment linking to the CMake docs here, so that others can quickly understand what this code does?

```suggestion
      # the following is used for integration with FindPackage
      # https://cmake.org/cmake/help/latest/module/FetchContent.html#fetchcontent-find-package-integration-examples
      string(TOLOWER ${Name} NameLower)
```

TheLartians (commented):
Thanks a lot for the PR and sorry for taking long to review it! It looks great on first glance, I have one small suggestion for readability. Also it looks like the CI pipelines are not running, do you need to enable GitHub actions for your fork?

patstew:
I have added some documentation.
I think you need to approve it to run the tests https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks

TheLartians:
> There should be an "Approve and Run" button somewhere for you?

Yep this time it showed the button, thanks for update!

Looks like we need to update the tests to support the different behavior based on the CMake version and also run
The code formatter (see [here](https://github.com/cpm-cmake/CPM.cmake/blob/master/CONTRIBUTING.md) for more info).

patstew:
I think I've fixed the issues. I've also added a few checks to the modules.cmake test so that it fails without the fixes in this commit, and ensures the fallback to the old method works. I've got rid of the simple check that FindA.cmake exists and reinstated the check that `find_package` works instead. Was there was a reason for the `find_package` to be disabled previously?

TheLartians (approved):
Looks great, thanks for updating this feature and the according documentation and test changes!

> Was there was a reason for the find_package to be disabled previously?

Probably an oversight, good catch!

TheLartians:
Changes released in `v0.40.3` 🎉

jay3d:
This change breaks my configuration that worked in `v0.40.2` before.
I have two packages that are added via CPM, and one is dependent on the other, now the "next" package won't find the first one anymore after this change, that was working fine before by default:

```cmake
CPMAddPackage(
        NAME TBB
        GIT_REPOSITORY https://github.com/oneapi-src/oneTBB.git
        GIT_TAG v2021.13.0
        EXCLUDE_FROM_ALL YES
        SYSTEM YES
)

CPMAddPackage(
        NAME OpenSubdiv
        GIT_REPOSITORY https://github.com/PixarAnimationStudios/OpenSubdiv.git
        GIT_TAG v3_6_0
        EXCLUDE_FROM_ALL YES
        SYSTEM YES
)
```

Now it can't find `TBB` anymore in `.3` and above.

PR #605, opened 2024-09-25
Update version checks to avoid cmake issue #26243

### LINKED ISSUE #603: GTest configuration fails as of CPM version ``0.40.1``
As the title already states, my project fails configuring when upgrading CPM to Version ``0.40.1`` and beyond, due to this error:
```cmd
1> [CMake] CMake Error: File /home/dnkpp/.vs/cpm_test/952dbcb6-53c6-4ed1-b2ba-4d91dcbd3971/out/build/Linux-Clang-Debug/_deps/gtest-src/cmake/Config.cmake.in does not exist.
1> [CMake] CMake Error at /usr/local/share/cmake-3.28/Modules/CMakePackageConfigHelpers.cmake:344 (configure_file):
1> [CMake]   configure_file Problem configuring file
1> [CMake] Call Stack (most recent call first):
1> [CMake]   /home/dnkpp/.vs/cpm_test/952dbcb6-53c6-4ed1-b2ba-4d91dcbd3971/out/build/Linux-Clang-Debug/_deps/gtest-src/googletest/CMakeLists.txt:102 (configure_package_config_file)
1> [CMake] -- Configuring incomplete, errors occurred!
```
This is just the output from my minimal example. If you want to have a look at my actual project output, here is the relevant gh-action run: https://github.com/DNKpp/mimicpp/actions/runs/10965745313
This happens to be the case for all linux  based machines, I've tested it (also some local docker containers).

I'll attach the minimal example: [cpm_test.zip](https://github.com/user-attachments/files/17081498/cpm_test.zip)

Avus-c:
Fixes #603  
Related to #570

This PR updates the CMake version check from `3.28.0` to `3.30.0` due to a known bug affecting versions up to `3.30.2`. The issue has been resolved in CMake `3.30.3` (see CMake [issue #26243](https://gitlab.kitware.com/cmake/cmake/-/issues/26243)).

Users may still encounter the issue described in #603 when using CMake versions `3.30.0` to `3.30.2`.

Instead of using the non-deprecated function starting from `3.30.0`, we could set it to `3.30.3`, where the bug is fixed.
However, it will then trigger deprecation warnings.

Basically if someone uses CPM >=0.40.1, and cmake 3.30.0 - 3.30.2, they run into either deprecation errors or a cmake bug, no matter what we do.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -869,8 +869,9 @@ function(CPMAddPackage)
     # Calling FetchContent_MakeAvailable will then internally forward these options to
     # add_subdirectory. Up until these changes, we had to call FetchContent_Populate and
     # add_subdirectory separately, which is no longer necessary and has been deprecated as of 3.30.
+    # A Bug in CMake prevents us to use the non-deprecated functions until 3.30.3.
     set(fetchContentDeclareExtraArgs "")
-    if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.28.0")
+    if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.30.3")
       if(${CPM_ARGS_EXCLUDE_FROM_ALL})
         list(APPEND fetchContentDeclareExtraArgs EXCLUDE_FROM_ALL)
       endif()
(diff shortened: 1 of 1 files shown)


cmake/CPM.cmake
```diff
@@ -870,7 +870,7 @@ function(CPMAddPackage)
     # add_subdirectory. Up until these changes, we had to call FetchContent_Populate and
     # add_subdirectory separately, which is no longer necessary and has been deprecated as of 3.30.
     set(fetchContentDeclareExtraArgs "")
-    if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.28.0")
+    if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.30.0")
```
gerhardol, line 873: ?
```suggestion
    if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.30.3")
```

Avus-c: fixed :)

TheLartians (approved):
Thanks for raising the issue and this fix! I agree that it's better to encounter a deprecated warning instead of the bug.

TheLartians:
Released in `v0.40.4` 🎉

PR #617, opened 2024-11-07
feature: allow URI to use shorthand syntax with additional options

SGSSGene:
(This is an alternative to PR #546 )
This allows to combine the shorthand syntax by using URI with additional arguments:

CPMAddPackage(URI "gh:nlohmann/json@3.9.1" OPTIONS "JSON_BUildTests OFF")

This is much shorter than the longer syntax way of writing:

CPMAddPackage (
  NAME nlohmann_json
  VERSION 3.9.1
  OPTIONS
    "JSON_BuildTests OFF"
)


test/integration/test_simple.rb
```diff
@@ -83,9 +83,39 @@ def test_update_single_package
...
+      # * using-adder - for this project
+      # ...and notably no test for adder, which must be disabled from the option override from above
+      assert_equal ['simple', 'using-adder'], exes
+
+    }
 
     create_with_commit_sha.()
     update_to_version_1.()
     update_with_option_off_and_build.()
+  update_with_option_off_and_build_with_uri_shorthand_syntax.()
```
TheLartians, line 119: For style consistency.

```suggestion
    update_with_option_off_and_build_with_uri_shorthand_syntax.()
```

README.md
```diff
@@ -111,6 +112,13 @@ CPMAddPackage("https://example.com/my-package-1.2.3.zip#MD5=68e20f674a48be38d60e
 CPMAddPackage("https://example.com/my-package.zip@1.2.3")
 ```
 
+Additionally, the shorthand syntax can be used with the URI and combined with the other options from above:
+```cmake
+CPMAddPackage("gh:nlohmann/json@3.9.1"
+  OPTIONS "JSON_BuildTests OFF"
+)
```
TheLartians, line 119: I think this needs to be updated for the named argument.

```suggestion
CPMAddPackage(
  URI "gh:nlohmann/json@3.9.1"
  OPTIONS "JSON_BuildTests OFF"
)
```

README.md
```diff
@@ -464,9 +466,7 @@ endif()
 
 ```cmake
 CPMAddPackage(
-  NAME lua
-  GIT_REPOSITORY https://github.com/lua/lua.git
-  VERSION 5.3.5
+  URI "gh:lua/lua@5.3.5"
```
TheLartians, line 469: I think it would be a good idea to keep at least one example with the old API so users can easily understand how it works.

```suggestion
  NAME lua
  GIT_REPOSITORY https://github.com/lua/lua.git
  GIT_TAG v5.3.5
```

TheLartians (commented):
Hey, thanks a lot for this alternative implementation and thanks for your patience, as I'm quite busy these days! I personally much prefer this alternative to the other PR, as it looks much cleaner and is not mixing different ways to pass arguments (I know it's a common CMake style, so this is a very subjective take). I was hoping to hear some input from @iboB on this, but I think he is very busy as well.

> Important difference to https://github.com/cpm-cmake/CPM.cmake/pull/546, it does not set EXCLUDE_FROM_ALL;YES;SYSTEM;YES;

Can you elaborate what the advantage is? I think I it would be intuitive for both `CPMAddPackage("gh:nlohmann/json@3.9.1")` and `CPMAddPackage(URI "gh:nlohmann/json@3.9.1")` to work the same way, so IMO both versions should set the same flags.

Sorry again for taking so long to respond. I have some more time during the holidays, so happy to help get this merged asap.

TheLartians:
> Can you elaborate what the advantage is?

I think I found the answer in the other PR:

> Harder to document. Currently we have two cases: 1. normal usage, 2. shorthand syntax. With URI we kind of have to split the 'normal usage' into two cases. I don't see how this can be documented in any understandable way. (Assuming we don't delete the 2. shorthand syntax)

I see your concern, though I feel it's quite easy to understand in the docs:

> All packages added using the shorthand syntax will be added using the [EXCLUDE_FROM_ALL](https://cmake.org/cmake/help/latest/prop_tgt/EXCLUDE_FROM_ALL.html) and [SYSTEM](https://cmake.org/cmake/help/latest/prop_tgt/SYSTEM.html#prop_tgt:SYSTEM) flag.
> ...
> Additionally, the shorthand syntax can be combined with the other options from above[ using the `URI` argument]:

SGSSGene:
PR #546 basic idea was, take the "short hand" syntax and allow extended options.
This PR took the standard approach and added a `URI` which is a short from for repo, name and version.
That is why I didn't touch EXCLUDE_FROM_ALL and SYSTEM.
But from what I understand, you would prefer a URI which allows a short form of repo, name and version + also sets EXCLUDE_FROM_ALL and SYSTEM.

I will adjust this PR accordingly

SGSSGene:
This new version automatically sets `EXCLUDE_FROM_ALL` and `SYSTEM`. I made sure that this can be overridden by setting `EXCLUDE_FROM_ALL OFF` or `SYSTEM OFF`. (And added a unit/integration test for this case).

I added another note below the `CPMAddPackage(....)` description noting the behavior of URI.

test/integration/test_simple.rb
```diff
@@ -83,9 +83,69 @@ def test_update_single_package
...
+      assert File.directory? exe_dir
+
+      exes = Dir[exe_dir + '/**/*'].filter {
+        # on multi-configuration generators (like Visual Studio) the executables will be in bin/<Config>
+        # also filter-out other artifacts like .pdb or .dsym
+        !File.directory?(_1) && File.stat(_1).executable?
+      }.map {
+        # remove .exe extension if any (there will be one on Windows)
+        File.basename(_1, '.exe')
+      }.sort
```
TheLartians, line 106: Not very important, but this code is now duplicated three times in the test case. Could it be more readable if we turn it into a method?

TheLartians (changes requested):
Thanks for the updates! I really like how it's turning out. There are some small stylistic and documentation issues left and then I'm happy to merge this!

Richardk2n:
Dear distinguished Gentleman,
how's it going?

README.md
```diff
@@ -112,6 +115,13 @@ CPMAddPackage("https://example.com/my-package-1.2.3.zip#MD5=68e20f674a48be38d60e
 CPMAddPackage("https://example.com/my-package.zip@1.2.3")
 ```
 
+Additionally, the shorthand syntax can be used with the long version, using the `URI` specifier.
```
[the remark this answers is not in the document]
SGSSGene: :+1:

README.md
```diff
@@ -65,6 +65,7 @@ Afterwards, any targets defined in the dependency can be used directly.
 
 ```cmake
 CPMAddPackage(
+  URI           # shorthand including repo, name, version and tag (see shorthand syntax)
```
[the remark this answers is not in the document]
SGSSGene: Sure!

cmake/CPM.cmake
```diff
@@ -620,10 +612,26 @@ function(CPMAddPackage)
...
+    # The shorthand syntax implies EXCLUDE_FROM_ALL and SYSTEM
+    set(ARGN "${ARGN};EXCLUDE_FROM_ALL;YES;SYSTEM;YES;")
+
+    # Parse URI shorthand argument
+  elseif(argnLength GREATER 1 AND "${ARGV0}" STREQUAL "URI")
+    list(REMOVE_AT ARGN 0 1) # remove "URI gh:<...>@version#tag"
+    cpm_parse_add_package_single_arg("${ARGV1}" ARGV0)
+
+    set(ARGN "${ARGV0};EXCLUDE_FROM_ALL;YES;SYSTEM;YES;${ARGN}")
+  endif()
```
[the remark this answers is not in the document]
SGSSGene: The challenge here is, that I rebuild the "ARGN" Variable, If URI is not the first argument it will get less straight forward to remove the it from ARGN. In that case I need to remove URI and move the <uri> information to the front.

test/integration/test_simple.rb
SGSSGene: I agree. Sadly my ruby skills are non-existent. So I rather touch this code as little as possible.

SGSSGene:
> Dear distinguished Gentleman, how's it going?

Thanks for the reminder!
I updated the PR.

TheLartians (commented):
Thanks for the changes, I like how it turned out! As you've allowed maintainer commits and there's not much left I'll add the final touches mentioned below and merge if tests pass. :)

TheLartians (approved):

TheLartians:
URI argument support is released in `v0.41.0` 🎉 
Thanks again the for PR and updates!

SGSSGene:
> Thanks for the changes, I like how it turned out! As you've allowed maintainer commits and there's not much left I'll add the final touches mentioned below and merge once tests pass. :)

Yeah! Thank you!!!

PR #622, opened 2024-11-21
CPM.cmake redundant return() removed

artiomn:
CPM can work, but this return stop to loading script and functions are unavailable.


TheLartians:
Hey, thanks for the PR! I'm not quite sure I understand implication of the change however. The intent of the early return is that if CPM is already loaded by a downstream dependency we don't overwrite the current CPM with another version and loose all state variables etc. Maybe it would be more clear if you can create a test case that fails with the current version?

artiomn:
Yes, and problem is here: some dependency uses CPM and my project uses CPM.
After loading dependency CPM doesn't initialized in the my CMakeLists.
And I can't use it in my project.

I've had this problem in the real project.
If it will be repeated, I'll try to make test example.

PR #627, opened 2024-12-29
Fix finding patch executable when cross-compiling on windows.

rokups:
`WIN32` checks whether current build context is windows. This is not correct in case of cross-compiling, because we are interested in finding host's tools, not SDK's tools.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -490,7 +490,7 @@ function(cpm_add_patches)
 
   # Find the patch program.
   find_program(PATCH_EXECUTABLE patch)
-  if(WIN32 AND NOT PATCH_EXECUTABLE)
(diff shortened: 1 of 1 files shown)


TheLartians:
Thanks for raising the issue and submitting a fix! For some reason I cannot get the CI to run. Could you check if you can enable GitHub Actions on your fork?

rokups:
Hmm somehow workflows do not start on my repo. I see workflows triggered on CPM.cmake repo, but github says a maintainer has to approve them. Someone has to click something somewhere..

TheLartians:
Ah I was using the "new merge experience" from GitHub and it was missing the button. Turns out switching to classic made it appear. Thanks!

TheLartians (approved):
Thanks for the fix!

TheLartians:
Fix deployed in `v.040.5` 🎉

Issue #629, opened 2025-01-05
This change #604 breaks my configuration that worked in `v0.40.2` before.


jay3d:
This change breaks my configuration that worked in `v0.40.2` before.
I have two packages that are added via CPM, and one is dependent on the other, now the "next" package won't find the first one anymore after this change, that was working fine before by default:

```cmake
CPMAddPackage(
        NAME TBB
        GIT_REPOSITORY https://github.com/oneapi-src/oneTBB.git
        GIT_TAG v2021.13.0
        EXCLUDE_FROM_ALL YES
        SYSTEM YES
)

CPMAddPackage(
        NAME OpenSubdiv
        GIT_REPOSITORY https://github.com/PixarAnimationStudios/OpenSubdiv.git
        GIT_TAG v3_6_0
        EXCLUDE_FROM_ALL YES
        SYSTEM YES
)
```

Now it can't find `TBB` anymore in `.3` and above.

_Originally posted by @jay3d in https://github.com/cpm-cmake/CPM.cmake/issues/604#issuecomment-2571353162_

patstew:
Can you see if #630 fixes it?

jay3d:
> Can you see if #630 fixes it?

Yes it does, thanks for looking into this issue!

PR #630, opened 2025-01-06
Fix version filename for find_package redirects

patstew:
I think I made a mistake copying what FetchContent does in #604 which causes #629.


patstew:
I don't really understand how that build error in sol2 could possibly be related to this PR. Has clang been updated, or the relevant tags in the upstream repo changed?

christopherbate:
Yes GitHub runner image `ubuntu-latest` was updated to 24.04 last month, after the last time that workflow ran successfully in this Repo. Probably need to set the runner to `ubuntu-22.04` on the `examples/clang` job

TheLartians (approved):
Thanks for fixing the issue! I'll try to find out what's up with sol2 and push changes here if needed.

TheLartians:
Fix released in `v0.40.6` 🎉

PR #648, opened 2025-03-02
Investigate sol2 CI build issue

TheLartians:
In seems that clang 18.1.0 creates an issue with sol2, leading to compilation errors. This PR adds a patch taken from https://github.com/ThePhD/sol2/issues/1581#issuecomment-2103463524 that fixes the issue.


TheLartians:
The error seems to be related to https://github.com/ThePhD/sol2/issues/1581

PR #652, opened 2025-03-04
Fix "cpm_add_patches" handling of multiple patch commands

christopherbate:
Update "cpm_add_patches" to correctly handle multiple patch files. The "PATCH_COMMAND" executor may not be a shell, so it is not portable to use shell operators like "&&" and "<" in the command. Commands after the initial PATCH_COMMAND should nstead be prefixed by COMMAND, and we can use "-i" to specify that a file is used by the "patch" command.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -529,16 +529,16 @@ function(cpm_add_patches)
     # Convert to absolute path for use with patch file command.
     get_filename_component(PATCH_FILE "${PATCH_FILE}" ABSOLUTE)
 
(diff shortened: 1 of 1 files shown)


christopherbate:
@TheLartians can you review or can you approve workflows?

TheLartians (changes requested):
Thanks for the ping and fix! The changes look fine, but may require running cmake-format (see the project;s [styleguide](https://github.com/cpm-cmake/CPM.cmake/blob/master/CONTRIBUTING.md#styleguides)). Also do you think you could add / modify a test case to ensure multiple patches work as expected?

PR #654, opened 2025-03-08
do not generate module files with DOWNLOAD_ONLY

### LINKED ISSUE #649: How to use `find_package` with a user provided Find* module?
I'm trying to download a project which does not contain a CMake configuration from GitHub using

```cmake
cpmaddpackage(
  NAME
  CWPack
  GITHUB_REPOSITORY
  "clwi/CWPack"
  GIT_TAG
  1.4
  SYSTEM
  YES
  DOWNLOAD_ONLY
  YES)
```

and then provide my own Find* module by adding a `FindCWPack.cmake` file  inside a `cmake` subdirectory.

However, trying to call `find_package` to find this file does not work.

```cmake
list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)

find_package(CWPack)
```

At least not without also setting `CPM_DONT_UPDATE_MODULE_PATH`.

Is this intended behavior? I think it is a bit confusing, specially if `DOWNLOAD_ONLY` is specified as well?

Avus-c:
Resolves #649

Prevents calling `cpm_create_module_file` when `DOWNLOAD_ONLY` was set to true. 
I see no reason to generate module files, when you're just downloading a dependency.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -867,7 +867,9 @@ function(CPMAddPackage)
     endif()
   endif()
 
-  cpm_create_module_file(${CPM_ARGS_NAME} "CPMAddPackage(\"${ARGN}\")")
+  if(NOT "${DOWNLOAD_ONLY}")
+    cpm_create_module_file(${CPM_ARGS_NAME} "CPMAddPackage(\"${ARGN}\")")
+  endif()
 
   if(CPM_PACKAGE_LOCK_ENABLED)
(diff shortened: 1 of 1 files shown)


TheLartians (approved):
Thanks for the fix! I see how this can be confusing behaviour.

Perhaps might make sense to add a comment in another PR so future editors understand the reason the exception.

PR #669, opened 2025-06-19
Allow users to override the default path for CPM_SOURCE_CACHE

blozano-tt:
Currently, only the ENV CPM_SOURCE_CACHE can be used for steering CPM download caching.

However, I am not a fan of environment variable engineering, (We have too much of it already in our code base).

With this change, a user of CPM can first set CPM_SOURCE_CACHE_DEFAULT, and have that be respected by CPM.cmake.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -155,7 +155,10 @@ set(CPM_DRY_RUN
     CACHE INTERNAL "Don't download or configure dependencies (for testing)"
 )
 
-if(DEFINED ENV{CPM_SOURCE_CACHE} AND NOT DEFINED CPM_SOURCE_CACHE_DEFAULT)
+# ENV gets top priority to control CPM_SOURCE_CACHE location
+# If it is not defined, allow user to control the location with CPM_SOURCE_CACHE_DEFAULT
+# If neither are set disable CPM_SOURCE_CACHE
+if(DEFINED ENV{CPM_SOURCE_CACHE})
   set(CPM_SOURCE_CACHE_DEFAULT $ENV{CPM_SOURCE_CACHE})
 elseif(NOT DEFINED CPM_SOURCE_CACHE_DEFAULT)
   set(CPM_SOURCE_CACHE_DEFAULT OFF)


blozano-tt:
@TheLartians does this PR sound okay? Just want to optionally steer this behavior without an environment variable. PS loving CPM over at @tenstorrent

iboB:
@blozano-tt you can configure with `-DCMP_SOURCE_CACHE="some/path"`. This however contrary to the PR takes precedence over an env var. If it's defined, the env var is ignored.

As a side note here's a piece of code that I use often:

```cmake
# before including CPM.cmake or get_cpm.cmake
if(NOT CPM_SOURCE_CACHE AND NOT DEFINED ENV{CPM_SOURCE_CACHE})
    set(CPM_SOURCE_CACHE "${CMAKE_SOURCE_DIR}/.cpm" CACHE PATH "CPM source cache")
    message(STATUS "Setting cpm cache dir to: ${CPM_SOURCE_CACHE}")
endif()
```

This allows me to cache `/repo-root/.cpm` on a CI, and reduces the download count for multiple config on dev machines which don't have a global setting.

iboB:
Ah. I think I get it now. The goal of this PR is not to make CPM_SOURCE_CACHE_DEFAULT something that one would configure with, but to reduce snippets like yours and mine from several lines to just:

```cmake
set(CPM_SOURCE_CACHE_DEFAULT "${CMAKE_SOURCE_DIR}/.cpm")
```

I'm not necessarily against that. However this does not affect just `CPM.cmake` itself, but also `get_cpm.cmake`. `get_cpm` respects the source cache and our snippets working on `CPM_SOURCE_CACHE` would affect it properly. Such behavior needs to be reflected there as well

blozano-tt:
> Hey, thanks for the PR!
> 
> In general, I'm not a fan of having multiple ways to achieve the same result, as it adds potential sources of confusion. Therefore, we need to be careful when considering extra options like this.
> 
> That said, with the current workarounds presented, I see an issue where an upstream dependency using the override could change the default cache location or suddenly enable the cache when reconfiguring a project. I think the suggested change would add a canonical way of setting the default cache location that doesn't have as many potential issues. So IMO it's a good idea to add it.
> 
> To get this merged I would request two additions:
> 
> * As @iboB mentioned, could you also update [`get_cpm.cmake`](https://github.com/cpm-cmake/CPM.cmake/blob/master/cmake/get_cpm.cmake) to check for `CPM_SOURCE_CACHE_DEFAULT` before falling back to no cache?
> * Could you document this option in the readme, so users can see how to set the default cache location? We might want to mention to not use locations outside of the source or build directory to avoid different projects creating different global caches on a system.
> 
> Thanks!

Will make the requested changes in a couple weeks after summer vacay!

sean-parent:
Just as an FYI - my default setup includes a test that the project is top-level to avoid enabling a cache if my client doesn't want one.

```cmake
# Setup CPM
if(PROJECT_IS_TOP_LEVEL AND NOT CPM_SOURCE_CACHE AND NOT DEFINED ENV{CPM_SOURCE_CACHE})
    set(CPM_SOURCE_CACHE "${CMAKE_SOURCE_DIR}/.cache/cpm" CACHE PATH "CPM source cache")
    message(STATUS "Setting cpm cache dir to: ${CPM_SOURCE_CACHE}")
endif()
include(cmake/CPM.cmake)
```

PR #681, opened 2025-11-09
Add Tracy to "Built with CPM.cmake"

gav2xlin:
Hello

This project is built with CPM.cmake [https://github.com/wolfpld/tracy/blob/master/cmake/vendor.cmake](https://github.com/wolfpld/tracy/blob/master/cmake/vendor.cmake)

I also corrected the links to other projects


TheLartians (approved):
Thanks for fixing the image URLs adding tracy! I wasn't aware it was using CPM. :)

PR #682, opened 2025-11-20
Support org-prefixed package names in find_package

sean-parent:
Change CPM.cmake to attempt to locate packages using organization-prefixed names (e.g., org-library) if the standard name is not found. This helps resolve naming conflicts and improves compatibility with packages installed under organization-qualified names. README documentation has been updated to reflect this behavior.


TheLartians (commented):
Thanks for the PR! I can see how this can be useful with `CPMFindPackage`. Two questions:
- I am currently not aware of this schema for org qualified package names. Is this a common practice / do you have a reference for it?
- I wonder about the ordering, as I would expect the org-qualified name to take priority over the unqualified name. Would it make sense to look for the org-qualified name first?

sean-parent:
The idea came to me while I was working on this CMake library built with CPM. https://github.com/stlab/cpp-library My workaround was to rename my repos with the prefixed name (i.e., https://github.com/stlab/copy-on-write became https://github.com/stlab/stlab-copy-on-write). This is acceptable, but it's not as clean as the old naming.

Issue #685, opened 2026-01-14
file download issues related to TLS


G4m4:
Hi!

On multiple repos relying on CPM for their dependencies I sometimes (on Windows computers only) struggle to download the content of the file CPM.cmake

The main symptom is this typical error during the configure step: `Unknown CMake command "CPMAddPackage"`

But it actually comes from the fact that the downloading of the file as done in get_cpm.cmake failed.
When retrieving the status of
```
  file(
    DOWNLOAD
    https://github.com/cpm-cmake/CPM.cmake/releases/download/v${CPM_DOWNLOAD_VERSION}/CPM.cmake
    ${CPM_DOWNLOAD_LOCATION}
    EXPECTED_HASH SHA256=${CPM_HASH_SUM}
    STATUS DOWNLOAD_STATUS
  )
```
`${DOWNLOAD_STATUS}` will contain `35;"SSL connect error"`

Although I have not investigating the root cause for the TLS issue as on WIndows it might become tricky.

PR #686, opened 2026-01-14
Workaround for #685: retry download failed due to TLS

G4m4:
In case of an "error 35" status ("SSL connect error"), retry downloading with TLS verify turned off.

We still expect the same hash.

### DIFF
--- a/cmake/get_cpm.cmake
+++ b/cmake/get_cpm.cmake
@@ -16,9 +16,31 @@ endif()
 # Expand relative path. This is important if the provided path contains a tilde (~)
 get_filename_component(CPM_DOWNLOAD_LOCATION ${CPM_DOWNLOAD_LOCATION} ABSOLUTE)
 
-file(DOWNLOAD
-     https://github.com/cpm-cmake/CPM.cmake/releases/download/v${CPM_DOWNLOAD_VERSION}/CPM.cmake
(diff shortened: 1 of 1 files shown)


TheLartians (commented):
Thanks for the PR! I'm not sure about this one, automatically retrying with disabled TLS verification does seem rather dodgy (we might as well just turn it off from the start).
Where does the error come from? If it's a proxy issue on your side, would it be a an option for you to just set [`CMAKE_TLS_VERIFY=OFF`](https://cmake.org/cmake/help/latest/envvar/CMAKE_TLS_VERIFY.html#envvar:CMAKE_TLS_VERIFY) on your system instead?

G4m4:
Yes, of course I understand that feeling, but the fact that the hash is still checked should prevent file tampering.

I have not been able to get to the root cause of this for my own case (and to be honest I do not think it is completely worth it considering the workaround is easy) but there are multiple examples of similar cases:
https://github.com/microsoft/vcpkg/issues/13368
https://gitlab.kitware.com/cmake/cmake/-/issues/24093

It all seems to relate to various combination of SSL library or softwares using it, and on Windows the typical "update your system" fix is not so easily done as a on e.g. Linux distros.

Of course if you feel like this is not a good idea we would indeed disable this locally, but some developer end users might get the same behaviour.

PR #687, opened 2026-01-19
Add support for GIT_SUBMODULES argument

gillamkid:
CPM does passthrough for many of the argument options of cmake's `FetchContent()` function. Hwever, there was an argument of interest that it was not passing: `GIT_SUBMODULES`.

`GIT_SUBMODULES` is needed if you want to use `CPMAddPackage()` and have it not initialize the submodules of the repo. 

I added support so `GIT_SUBMODULES ""` can be used to prevent submodules from being initialized the same way it is used to do so when calling `FetchContent()` directly.

### DIFF
--- a/cmake/.cmake-format-additional_commands-cpm
+++ b/cmake/.cmake-format-additional_commands-cpm
@@ -21,6 +21,7 @@ parse:
         FIND_PACKAGE_ARGUMENTS: 1
         NO_CACHE: 1
         GIT_SHALLOW: 1
+        GIT_SUBMODULES: 1
         URL: 1
         URL_HASH: 1
         URL_MD5: 1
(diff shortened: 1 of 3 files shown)


TheLartians:
Thanks for raising the issue! We should definitely support this edge case. Three things
- Can you add a test case so we don't break this accidentally?
- The issue is not specific to `GIT_SUBMODULES` issue will happen any time someone tries to pass an empty string to a forwarded parameter. Do you think there's a way we can make this a more general fix? E.g. writing a wrapper for `cmake_parse_arguments` that ensures empty arguments are preserved?
- Please run the code formatter (see [CONTRIBUTING.md](https://github.com/cpm-cmake/CPM.cmake/blob/master/CONTRIBUTING.md))

TheLartians:
@gillamkid note that this issue is also handled by https://github.com/cpm-cmake/CPM.cmake/pull/461, which resolves the issue in a more general way. The PR has gone stale due to lack of time from my side, but given the obvious need for a fix I will prioritise it.

gillamkid:
I added some unit tests and fixed the styling. @TheLartians can you please approve the Gitlab Actions workflow to run?

PR #655, opened 2025-03-13
fix the warning CPM: Your project is using an unstable development version of CPM.cmake. when using FetchContent

flagarde:
This PR fix the warning 
```bash
CMake Warning at build/_deps/cpm-src/cmake/CPM.cmake:80 (message):
  CPM: Your project is using an unstable development version of CPM.cmake.
  Please update to a recent release if possible.  See
  https://github.com/cpm-cmake/CPM.cmake for details.
Call Stack (most recent call first):
  build/_deps/cpm-src/CMakeLists.txt:37 (include)
```

when using FetchContent to download CPM.

describe --tags --long command : 
Always output the long format (the tag, the number of commits and the abbreviated commit name) even when it matches a tag. This is useful when you want to see parts of the commit object name in "describe" output, even when the commit in question happens to be a tagged version. Instead of just emitting the tag name, it will describe such a commit as v1.2-0-gdeadbee (0th commit since tag v1.2 that points at object deadbee…​.).
 
so if the HEAD is exactly at some tags matching v*.*.*-0-* then Head is on a release and so not a development.

PR #691, opened 2026-04-12
fix CPMs warning message: Your project is using an unstable development version of CPM.cmake, when the project is git cloned

marrinosnis:
This PR fixes the same issue as [this PR](https://github.com/cpm-cmake/CPM.cmake/pull/655), when the cpm-cmake project is **cloned** and is included in the main project of the user by him. The problem seems to be that at the CMakeList.txt file, at the REGEX expression a hardcoded `0` was placed in the expression. Now, where the version changed to `1`, the comparison 

```cmake
if(DESCRIBE_NAME STREQUAL "")
  set(CPM_DEVELOPMENT "-development-version")
endif()
``` 
always ends up to true. 

The variable `DESCRIBE_NAME` is empty, so finally the `EXTRACTED_CPM_VERSION` that is calculated later at the same file
```cmake
string(REGEX MATCH "([0-9\\.]+)" EXTRACTED_CPM_VERSION "${DESCRIBE_NAME}")
```
and is used at the CPM.cmake, always has the value `development-version`, resulting to showing that warning message.

Also, small change added to CPM.cmake file, in order to cover the case where the CPM project cloned, instead of used any of mentioned ways (using FetchContent or by running get_cpm.cmake). Probably someone, just like me, would like to test the external project first, by git cloning it, and then added to his personal project.

The changes at the cmake/CPM.cmake file, aim to `add_subdirectory()`, if the CPM is cloned instead of downloading it using the previous mentioned ways.


marrinosnis:
@iboB @TheLartians Hi. Are there any updates for this pull request?

cmake/CPM.cmake
```diff
@@ -42,8 +42,13 @@ if(NOT COMMAND cpm_message)
   endfunction()
 endif()
 
+get_filename_component(CURRENT_FILENAME_PATH "${CMAKE_CURRENT_LIST_FILE}" DIRECTORY)
+set(CPM_CMAKE_DIRECTORY "${CURRENT_FILENAME_PATH}/..")
+
 if(DEFINED EXTRACTED_CPM_VERSION)
   set(CURRENT_CPM_VERSION "${EXTRACTED_CPM_VERSION}${CPM_DEVELOPMENT}")
+elseif(EXISTS "${CPM_CMAKE_DIRECTORY}/CMakeLists.txt")
+  add_subdirectory(${CPM_CMAKE_DIRECTORY})
```
TheLartians, line 51: It seems to me like this could easily lead to unexpected behaviour or errors, e.g. if the CPM.cmake is explicitly included in a CMake project.

CMakeLists.txt
```diff
@@ -7,8 +7,10 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git/")
     WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
     OUTPUT_VARIABLE DESCRIBE_NAME COMMAND_ERROR_IS_FATAL ANY
   )
+
   string(STRIP ${DESCRIBE_NAME} DESCRIBE_NAME)
-  string(REGEX MATCH "v[0-9]*.[0-9]*.[0-9]*\-0\-[a-zA-Z0-9]*" DESCRIBE_NAME ${DESCRIBE_NAME})
+  string(REGEX MATCH "v[0-9]*.[0-9]*.[0-9]*\-[0-9]\-[a-zA-Z0-9]*" DESCRIBE_NAME ${DESCRIBE_NAME})
```
TheLartians, line 12: Note -0- is here to indicate that the current commit is indeed a tagged commit corresponding to a version (see also the original PR). Accepting other values would falsely identify the next 9 commits as releases.

TheLartians (commented):
Hey @marrinosnis, thanks for the PR and sorry for not being very active in this project. If I understand correctly, the warning you are seeing occurs because the cloned version of CPM.cmake is not a released version but on a commit sometime after a recent release, therefore triggering the warning. This is the intended behaviour of the warning and should not be changed.

marrinosnis:
@TheLartians  Please, let me write down the steps to reproduce the warning message:

1. Clone the CPM project based on the latest release tag. Currently is `v42.0.0` (command to achieve this: 
`git clone --depth 1 --branch v0.42.2 git@github.com:cpm-cmake/CPM.cmake.git`. This might leave you with detached HEAD, but doesn't really matter for our case).

2. Move the CPM project, inside the src directory of the main project you are working with, e.g. testCPM
`mv CPM.cmake testCPM/src/`
Probably the state of the src directory would be something like the following
image

3. Content of main CMakeLists.txt from the main project would be very simple (for testing purposes) like following:
```cmake
...

list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CPM.cmake/cmake")

include(CPM)

CPMAddPackage(
        GITHUB_REPOSITORY "example/gitRepository"  # download this external Git repository
        SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/myLib"  # create directory named 'myLib' and put it there
        GIT_TAG f6f22c59bb727895ff1a214b4565a1e595ec590  # download from this specific Git commit
)

add_executable(mainExec main.cpp)
...

``` 

The above steps finally will download the external project from the git repository, but will show the below warning message:
image

Again, just to make it clear, I don't want to have the main project checking if the CPM project exists, and if not download it. The steps that I am following are that I have **already** download the latest release version of the CPM project, and I have moved it inside my main project, which produces this warning.

cmake/CPM.cmake
marrinosnis: Please, check my below comment for explanation of the exacts steps to reproduce the case. Moreover, having added the CPM.cmake into a main project didn't cause and problems.

TheLartians:
Hey @marrinosnis thanks for providing the use case! I'm afraid that this will be hard to support without major changes to how the version is injected into the CPM.cmake script. For now I recommend using a script with version info as found in the [releases page](https://github.com/cpm-cmake/CPM.cmake/releases) or using `FetchContent` to include the CPM script in your projects.

marrinosnis:
Okay. Thanks for the note

PR #664, opened 2025-06-03
fix: Patch commands fail trying to re-apply the same patch twice

BohdanBuinich:
# Fix repeated patch application

## Problem

`PATCHES` used to build a shell-style `PATCH_COMMAND` that applied patch files with input redirection.

That made patching fragile:

- repeated configure runs could try to apply the same patch again;
- patch paths with spaces were harder to handle safely;
- CPM had no explicit record that a patch set had already been applied;
- changing a patch file at the same path could still look like the same patch command.

## What changed

Patch handling now generates a small CMake script and uses it as the `PATCH_COMMAND`.

The script applies patches from the dependency source directory without shell redirection. For each patch, it first checks whether the patch can apply normally. If not, it checks whether the patch is already present by running a reverse dry-run.

This makes repeated configure runs safe:

- apply the patch if it is not applied yet;
- skip it if the same patch set was already completed;
- accept it if the source tree already contains the patch changes;
- fail clearly if the patch is neither applicable nor already applied.

## Main idea

Patch state is now stored next to the dependency source tree:

```text
.cpm_patches/<patch_set_hash>.stamp
```

The patch-set hash includes the patch executable, patch arguments, working-directory behavior, patch file paths, and patch file contents.

That means changing a patch file creates a new patch set, even if the file path stays the same.

Fixes #618

Issue #618, opened 2024-11-18
[BUG] Patch commands fail trying to re-apply the same patch twice


Jazzinghen:
Due to a dependency with improper CMakeLists.txt I tried to use the `PATCHES` field, however the second time I run the configuration CMake fails during `FetchContent_MakeAvailable` for that package with this output:

```shell
[cmake] -- CPM: Adding package sqlite_protobuf@0 (082596c282cab90f50b6d6dee5f9be22fd503dbe)
[cmake] [ 11%] Performing update step for 'sqlite_protobuf-populate'
[cmake] -- Already at requested ref: 082596c282cab90f50b6d6dee5f9be22fd503dbe
[cmake] [ 22%] Performing patch step for 'sqlite_protobuf-populate'
[cmake] patching file CMakeLists.txt
[cmake] Reversed (or previously applied) patch detected!  Assume -R? [n] 
[cmake] Apply anyway? [n] 
[cmake] Skipping patch.
[cmake] 1 out of 1 hunk ignored -- saving rejects to file CMakeLists.txt.rej
[cmake] patching file src/CMakeLists.txt
[cmake] Reversed (or previously applied) patch detected!  Assume -R? [n] 
[cmake] Apply anyway? [n] 
[cmake] Skipping patch.
[cmake] 2 out of 2 hunks ignored -- saving rejects to file src/CMakeLists.txt.rej
[cmake] make[2]: *** [CMakeFiles/sqlite_protobuf-populate.dir/build.make:118: sqlite_protobuf-populate-prefix/src/sqlite_protobuf-populate-stamp/sqlite_protobuf-populate-patch] Error 1
[cmake] make[1]: *** [CMakeFiles/Makefile2:87: CMakeFiles/sqlite_protobuf-populate.dir/all] Error 2
[cmake] make: *** [Makefile:91: all] Error 2
[cmake] 
[cmake] CMake Error at /usr/share/cmake/Modules/FetchContent.cmake:1918 (message):
[cmake]   Build step for sqlite_protobuf failed: 2
[cmake] Call Stack (most recent call first):
[cmake]   /usr/share/cmake/Modules/FetchContent.cmake:1609 (__FetchContent_populateSubbuild)
[cmake]   /usr/share/cmake/Modules/FetchContent.cmake:2145:EVAL:2 (__FetchContent_doPopulation)
[cmake]   /usr/share/cmake/Modules/FetchContent.cmake:2145 (cmake_language)
[...]

PR #692, opened 2026-04-22
fix: PATCHES command when using package-lock

OptionalValue:
This pull-request is related to #618 and #664.

I've had several issues when trying to use CPM to apply a patch to a sub-dependency.
Every library uses package-lock files.

The dependency tree looks something like this:

```
Main project
└── Sub library
    └── External dependency that needs patching
    └── dependency.patch
```

When first trying this setup, I had the issue that the dependency.patch was applied multiple times, leading to the error described in #618. I've included the fix from #664 in this PR as it wont work without it.

The second issue was, that the sub library had a package lock file like this:
```
# sub-library
CPMDeclarePackage(sub-library
  GIT_TAG v1.0.0
  GIT_REPOSITORY xxx
  PATCHES
  "externals/dependency.patch"
)
```
So the patch file path was relative to the sub-library source dir.
However when running cpm-update-package-lock the file path was saved as an absolute path, which is not portable...
While fixing this, I noticed another problem:
the package-lock file of the main project gets the transient dependency and therefore also the PATCHES command from the sub-library, which is whats expected. However, now the CMake context is that of the main project, so we can't use any of the usual CMAKE_SOURCE_DIR variables etc in the path. This caused the patch file to not be found at all.
The only way I could think of to fix this was the following:

When running in the original sub-library context, the path is resolved (and found) and when no CMake variable is detected "${}", the current project source dir gets prepended to the path. This way the path is always relative to the actual target that contains the patch file.
I also fixed it, so that cpm-update-package-lock does not save the resolved absolute path but rather the unresolved path including all CMake variables before replacing them.

In the end this is the package-lock file which actually worked so far for me:

```
# sub-library
CPMDeclarePackage(sub-library
  GIT_TAG v1.0.0
  GIT_REPOSITORY xxx
  PATCHES
 "\${sub-library_SOURCE_DIR}/externals/dependency.patch"
)
```
Note: the CMake variable needs to be escaped, but this is done by CPM.
The user only needs to enter the relative path and when running cpm-update-package-lock it will preprend the SOURCE_DIR variable automatically.

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -1387,6 +1387,11 @@ function(cpm_prettify_package_arguments OUT_VAR IS_IN_COMMENT)
       if(${IS_IN_COMMENT})
         string(APPEND PRETTY_OUT_VAR "#")
       endif()
+      if(${oneArgName} STREQUAL "SOURCE_DIR")
+        string(REPLACE ${CMAKE_SOURCE_DIR} "\${CMAKE_SOURCE_DIR}" CPM_ARGS_${oneArgName}
+                       ${CPM_ARGS_${oneArgName}}
+        )
+      endif()
       string(APPEND PRETTY_OUT_VAR "  ${oneArgName} ${CPM_ARGS_${oneArgName}}\n")
     endif()
   endforeach()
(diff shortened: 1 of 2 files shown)


BohdanBuinich:
@OptionalValue I pushed my latest changes to https://github.com/cpm-cmake/CPM.cmake/pull/664, and it should also fix your issues. Could you please confirm that it works for you?

PR #694, opened 2026-04-29
Support .tar.xz and .tar.zst in tarball URLs, fix #536

### LINKED ISSUE #536: Pick up version from tarball name NAME-VERSION.tar.xz
I'm adding libogg like this:
```cmake
CPMAddPackage(NAME libogg
	URL http://downloads.xiph.org/releases/ogg/libogg-1.3.5.tar.xz
	URL_HASH SHA256=c4d91be36fc8e54deae7575241e03f4211eb102afb3fc0775fbbc1b740016705
	OPTIONS "BUILD_TESTING OFF")
```
which results in the following output:
```
CPM: Adding package libogg@ ()
```
For other downloads CPM automatically picks up the version number from the filename, would be great if it works here, too :)

jhasse:
see issue #536

### DIFF
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -258,7 +258,7 @@ endfunction()
 
 # Try to infer package name and version from a url
 function(cpm_package_name_and_ver_from_url url outName outVer)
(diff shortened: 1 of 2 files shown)


TheLartians (approved):
Thanks for adding support for these extensions!

PR #702, opened 2026-05-07
Clarify DOWNLOAD_ONLY option

ventZl:
Documentation wording suggests that DOWNLOAD_ONLY is a flag, whose presence is sufficient to disable project configuration. In fact DOWNLOAD_ONLY is an option taking one argument.


TheLartians (approved):
That's more clear, thanks!

PR #705, opened 2026-05-28
Add Windows SHA256 hash calculation instructions

studoot:
Added instructions for calculating SHA256 hash on Windows using PowerShell.


TheLartians (approved):
Thanks, very useful!

PR #707, opened 2026-06-09
Add GENERATED package lock mode (pnpm/npm style, commit-pinned)

jamierpond:
## Motivation

The current package lock records each dependency's declaration **verbatim** and only updates when you build the `cpm-update-package-lock` target. For projects that declare deps against a moving ref (e.g. `GIT_TAG main`), this has two rough edges compared to `pnpm-lock.yaml` / `package-lock.json`:

1. The lock isn't reproducible — it records `GIT_TAG main`, not the commit that was actually used.
2. It doesn't regenerate on its own. Delete it and reconfigure and it stays gone until you remember the separate target step.

## What this adds

A `GENERATED` keyword on `CPMUsePackageLock` (and an equivalent `CPM_GENERATE_PACKAGE_LOCK` option / env var):

```cmake
CPMUsePackageLock(package-lock.cmake GENERATED)
```

In this mode:

- **Auto-maintained**: the lock at the given path is rewritten in the source tree on every configure — no `cpm-update-package-lock` step. Delete it and reconfigure to refresh the pins, just like deleting a pnpm lock.
- **Commit-pinned**: git packages are pinned to the exact commit that was checked out, not the declared (possibly moving) `GIT_TAG`. The pin is then consumed on the next configure (declarations override call-site args), so it's stable/idempotent: `main` → resolved commit on the first run, identical lock thereafter.

Existing (non-`GENERATED`) behaviour is unchanged.

## Implementation notes

- The lock entry is now written **after** the fetch instead of before it, so the package source dir exists and its commit can be resolved. This is the only behavioural move for the existing path; the same set of packages is recorded as before.
- New helper `cpm_get_git_commit_hash()` reads the checked-out `HEAD` (falls back gracefully — no git / not a repo / URL packages → unchanged args).
- `CPMUsePackageLock` now `cmake_parse_arguments` for the `GENERATED` flag, includes any existing lock first (to apply current pins), then redirects `CPM_PACKAGE_LOCK_FILE` at the source path and rewrites it.

## Tests

Adds `test/unit/package-lock-generated.cmake` — hermetic (spins up a throwaway local git repo, **no network**) — asserting that after a plain configure (no target build) the lock is written into the source tree, pins the dependency to the resolved commit (and no longer references the branch), and is byte-identical across a reconfigure.

Formatted with the repo's `cmake-format` 0.6.11 config; existing package-lock unit tests still pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,6 +277,31 @@ cmake -Bbuild
 cmake --build build --target cpm-update-package-lock
 ```
 
+### Generated package lock (pnpm/npm style)
+
+Pass `GENERATED` to keep the lock in sync automatically, the way `pnpm-lock.yaml` or
+`package-lock.json` work:
+
+```cmake
+CPMUsePackageLock(package-lock.cmake GENERATED)
+```
+
+In this mode the lock at the given path is rewritten in your source tree on **every configure**,
+and git packages are pinned to the **exact commit that was checked out** rather than the (possibly
+moving) `GIT_TAG`. This makes the lock reproducible even when dependencies are declared against a
+branch like `main`. No separate `cpm-update-package-lock` step is needed.
+
+To refresh the pins (e.g. to pull in newer upstream commits), delete the lock file and reconfigure,
+just like deleting a `pnpm-lock.yaml`:
+
+```bash
+rm package-lock.cmake
+cmake -Bbuild
+```
+
+The same behaviour can be enabled globally without editing `CMakeLists.txt` by setting the
+`CPM_GENERATE_PACKAGE_LOCK` option (or environment variable).
+
 See the [wiki](https://github.com/cpm-cmake/CPM.cmake/wiki/Package-lock) for more info.
 
 ## Private repositories and CI
--- a/cmake/CPM.cmake
+++ b/cmake/CPM.cmake
@@ -129,6 +129,11 @@ option(CPM_INCLUDE_ALL_IN_PACKAGE_LOCK
        "Add all packages added through CPM.cmake to the package lock"
        $ENV{CPM_INCLUDE_ALL_IN_PACKAGE_LOCK}
 )
+option(
+  CPM_GENERATE_PACKAGE_LOCK
+  "Keep the package lock in sync on every configure (pnpm/npm style), pinning git deps to the resolved commit"
+  $ENV{CPM_GENERATE_PACKAGE_LOCK}
+)
 option(CPM_USE_NAMED_CACHE_DIRECTORIES
        "Use additional directory of package name in cache on the most nested level."
        $ENV{CPM_USE_NAMED_CACHE_DIRECTORIES}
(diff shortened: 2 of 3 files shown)


cmake/CPM.cmake
```diff
@@ -1047,6 +1077,35 @@ macro(CPMDeclarePackage Name)
...
+function(cpm_get_git_commit_hash repoPath result)
+  set(${result}
+      ""
+      PARENT_SCOPE
+  )
+
+  find_package(Git QUIET)
+  if(NOT GIT_EXECUTABLE OR NOT EXISTS "${repoPath}/.git")
+    return()
+  endif()
```
TheLartians, line 1091: In case of nested CPM dependencies, the value of `${result}`  may still be set in the parent scope leading to an incorrect resolution, so we should unset ${result} in the parent scope here.

cmake/CPM.cmake
```diff
@@ -1063,19 +1122,49 @@ function(cpm_add_comment_to_package_lock Name)
   endif()
 endfunction()
 
-# includes the package lock file if it exists and creates a target `cpm-update-package-lock` to
-# update it
+# Includes the package lock file if it exists and creates a target `cpm-update-package-lock` to
+# update it.
```
TheLartians: The docstring is no longer up-to-date: the `cpm-update-package-lock` target is no longer created when `GENERATED` or `CPM_GENERATE_PACKAGE_LOCK` is set.

cmake/CPM.cmake
```diff
@@ -1063,19 +1122,49 @@ function(cpm_add_comment_to_package_lock Name)
...
+          CACHE INTERNAL ""
       )
+      set(CPM_PACKAGE_LOCK_FILE
+          "${CPM_ABSOLUTE_PACKAGE_LOCK_PATH}"
+          CACHE INTERNAL ""
+      )
+      file(
+        WRITE ${CPM_PACKAGE_LOCK_FILE}
+        "# CPM Package Lock\n# This file is generated by CPM.cmake and should be committed to version control\n\n"
+      )
```
TheLartians, line 1168: It's a bit unclear to me what will happen if a dependency also uses CPM with a package lock file, I think this might overwrite the existing package lock again. I think calling `CPMUsePackageLock` should be a no-op if it has already been called at any point before.

README.md
```diff
@@ -277,6 +277,31 @@ cmake -Bbuild
...
+To refresh the pins (e.g. to pull in newer upstream commits), delete the lock file and reconfigure,
+just like deleting a `pnpm-lock.yaml`:
+
+```bash
+rm package-lock.cmake
+cmake -Bbuild
+```
+
+The same behaviour can be enabled globally without editing `CMakeLists.txt` by setting the
+`CPM_GENERATE_PACKAGE_LOCK` option (or environment variable).
```
TheLartians, line 310: Thanks for adding documentation! We should add that also changing any additional arguments like the `OPTIONS` will not work without removing the lockfile or the entry from the lockfile first.

Maybe it would also make sense to mark this as experimental for now, as for a proper npm-style lock one might expect.

- Changing the version or explicitly setting a commit hash in the source would also update the locked hash accordingly
- Any other flags from `CPMAddPackage` would still be respected, the lock should only impact the `GIT_TAG` and `GIT_REPOSITORY` entries.

Implementing this in the future could be a breaking change or lead to unexpected updates, so it makes sense to advertise caution for now.

TheLartians (changes requested):
Thanks for this PR, automatically locking the current commit hash is a big step for CPM's supply chain security and reproducibility - especially when moving targets are specified! I have some thoughts regarding using a auto-updating package lock in nested CPM dependencies and documentation, but looks great otherwise!

jamierpond:
@TheLartians thanks so much for looking at this! 

Meant for it to be marked as draft initially, am very grateful for your feedback.

Thanks again for giving this serious consideration, will revise with your feedback in mind!

PR #712, opened 2026-07-19
fix: Enhance CPMAddPackage with source override to preserve semicolons

### LINKED ISSUE #447: The Boost example does not work with  CPM_USE_LOCAL_PACKAGES=YES
If I try the `CPM.cmake v0.37.0` this is the result if the **Boost Libraries** are already installed on my build system:

```bash
-- CPM: Using local package Boost@1.81.0
-- Configuring done
CMake Error at CMakeLists.txt:21 (target_link_libraries):
  Target "CPMExampleBoost" links to:

    Boost::asio

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



-- Generating done
```

mathisloge:
The package-override test never tested the option preservation since the recursive call of CPMAddPackage when overriding the local source strips the list escapes. 

Fixes #447


TheLartians (dismissed):
Thanks a lot for the fix!

TheLartians (approved):

